Issue #9, opened 2015-02-13
css for SQL lessons


orchid00:
Hi, 
I have some suggestions on the template for the SQL lessons:

Padding:
*I see that most if not all tables need a bit more of bottom padding, the letters in the next paragraph are too close.
*Challenges on the other side are very spacious in between, and probably the question title should be a bit smaller.
Colors:
*Table headers, I don't know if it is just me, but that blue is too dark to be combined with black... just saying...

This is a duplicate from:
https://github.com/swcarpentry/sql-novice-survey/issues/27
suggested by Raniere Silva to be copied here.

pbanaszkiewicz:
Hi, @orchid00. I can't find challenges in the repository with lesson on SQL. Can you point me at them?

gvwilson:
http://github.com/swcarpentry/sql-novice-survey has a bunch of challenges.

pbanaszkiewicz:
@gvwilson do you both mean "FIXME"s?

orchid00:
I meant the font of the titles called "challenges" so all of them. It is a matter of style: they are very spacious in between, and probably the question title (where it says fixme) should be a bit smaller. Is that clearer?

pbanaszkiewicz:
Thanks @orchid00 and @gvwilson. My fork, for reason unknown to me, contained "FIXME" headers instead of proper challenges as, for example, here http://swcarpentry.github.io/sql-novice-survey/01-select.html (at the bottom).

I've pulled in the recent changes and will work on styles some more.

pbanaszkiewicz:
Hi everyone,

I don't like that light-green color of challenge boxes. I'd get rid of it, and to highlight that this section is a challenge I'd put it under "Challenge" (or "Exercise") header.

BEFORE:
![screenshot from 2015-02-14 12 12 50](https://cloud.githubusercontent.com/assets/72821/6199417/d61bed0c-b442-11e4-93b0-d1c508cd5bff.png)

AFTER:
![screenshot from 2015-02-14 12 12 17](https://cloud.githubusercontent.com/assets/72821/6199418/d99f6594-b442-11e4-9ecf-2bfd4b7b54b5.png)

Pros:

* clear indicator that this section is intended as an exercise for reader
* proper heading structure (each challenge is a subsection of "Challenges")
* visually more appealing

Cons:

* requires changes to lesson files

What do you think?

orchid00:
Good idea to remove the green boxes!
[...]

PR #10, opened 2015-02-13
Fix: styles for tables in SQL lessons

pbanaszkiewicz:
This PR addresses #9.

### Previously

* table cells had no padding, which caused a visual collapse (cell content was too close to cell neighbors)
* table header rows had dark blue background and black letters
* tables had no space between them and following paragraphs.

BEFORE:
screenshot from 2015-02-13 20 46 42

### Changes

* added 5px padding to cell contents
* changed header row background to gray-ish (I considered switching text color to white, but it didn't look good)
* 15px margin on the bottom of the tables (similarly to bottom margin used by paragraph style).

AFTER:
screenshot from 2015-02-13 20 51 07

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -466,17 +466,31 @@ section.content {
     font-size: 40pt;
 }
 
-thead .header {
-    background: rgba(0, 0, 0, 0.15);
+table {
+    margin-bottom: 15px;
 }
 
 table th, table td {
-    padding: 5px;
+    padding: 5px 10px;
 }
 
-table {
-    border: 1px solid rgba(0, 0, 0, 0.15);
-    margin-bottom: 15px;
+table > thead > .header {
+    background: transparent;
+}
+
+table > thead > tr > td, table > thead > tr > th,
+table > tbody > tr > td, table > tbody > tr > th,
+table > tfoot > tr > td, table > tfoot > tr > th {
+    border: 1px solid #DDD;
+}
+
+table > thead > tr > th,
+table > thead > tr > td {
+    border-bottom-width: 2px;
+}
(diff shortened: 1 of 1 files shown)


r-gaia-cs:
@pbanaszkiewicz What about use the same style from the table at our workshop template, see https://swcarpentry.github.io/workshop-template/? **You should keep the change for the table head**.

pbanaszkiewicz:
@r-gaia-cs thanks for suggestion. I was considering a switch to Bootstrap styles, but I don't entirely like them:

1) they have huge horizontal padding
2) they tinker with `:last-child` and `:first-child` a lot

If you have specific styles in mind that we can borrow (big padding, no background for `<thead>`, striped rows) please let me know and I'll implement it.

r-gaia-cs:
@pbanaszkiewicz I'm OK with the currently horizontal padding and agree to avoid playing a lot with `:last-child` and `:first-child`. What I'm suggesting is that we use

~~~
tbody > tr:nth-of-type(2n+1) {
    background-color: #F9F9F9;
}
~~~

and

~~~
tbody > tr > td {
    border-top: 1px solid #DDD;
}
~~~

**This samples need improvements to be more robust.**

pbanaszkiewicz:
Hi @r-gaia-cs,

this is what I have now:
screenshot from 2015-02-14 11 45 21

It's matching closely `.table-bordered` from Bootstrap. I decided to leave the header row background blank because it was too close in color to `#F9F9F9` - our rule for even rows in the table.

I also increased horizontal padding by 5px on both sides, because now that we have vertical borders it was too cramped.

PR #31, opened 2015-05-10
updating deck.js mostly to allow Alt+leftarrow to use the back action from the browser

twitwi:


gvwilson:
Taking deck.js support out (nobody's updating/maintaining slideshows).

PR #32, opened 2015-06-14
CSS cataloguing tool

gvwilson:
Add a tool to list the CSS styles used in a set of documents.


tools/catalog.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+
+'''Create YAML catalog of CSS styles using in a set of HTML documents.
```
kynan: used

PR #35, opened 2015-10-29
Drop to collapsed menu to prevent the menu dissapearing behind the logo.

Cadair:
replaces swcarpentry/site#1147.

Fixes this bug:

screen shot 2015-10-29 at 15 03 33

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -733,7 +733,7 @@ div.chapter h2 {
 }
 
 /* Collapse navbar */
-@media (max-width: 993px) {
+@media (max-width: 1200px) {
     .navbar-header {
         float: none;
 	min-height: 80px;


Cadair:
@fmichonneau The icon idea would be easy enough, but what would you want to happen when the button was clicked. That would be the bit that would be tricky.

fmichonneau:
I was thinking it would show the search bar. In other words, it would be a simplified version of the collapsed menu, but I might be overthinking this

Cadair:
@fmichonneau do you mean below the menu or overlaying on top of the menu?

fmichonneau:
I was thinking below the menu, but overlaying would work too...

Cadair:
@fmichonneau I like the idea, but I think it might push my CSS/JS skills to far.

PR #38, opened 2016-03-24
Use secure links to available external sites.

QuLogic:
Many of these redirect automatically, too.

See swcarpentry/workshop-template#312

### DIFF
--- a/css/bootstrap/bootstrap-js/bootstrap.js
+++ b/css/bootstrap/bootstrap-js/bootstrap.js
@@ -1,5 +1,5 @@
 /*!
- * Bootstrap v3.3.4 (http://getbootstrap.com)
+ * Bootstrap v3.3.4 (https://getbootstrap.com)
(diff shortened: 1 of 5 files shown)


rgaiacs:
This is from Bootstrap. We shouldn't change them.

QuLogic:
Sure, you can ignore the Bootstrap stuff; they're just comments. But `slideshow.css` and `swc.css` are not from Bootstrap. The links will cause mixed content warnings, or be ignored entirely.

rgaiacs:
@QuLogic only `css/slideshow.css` will cause content warnings. And slides were or will be deprecated.

wking:
On Sat, Mar 26, 2016 at 11:25:44AM -0700, Raniere Silva wrote:
> And slides were or will be deprecated.

I vaguely remember discussion (from January?) about reviving slides
for showing explanatory graphics when a second projector is available
(with the main projector still being used for live coding).  If I can
dig up a reference, I'll post it here.

PR #64, opened 2016-06-23
Only try building R files if there are some

gvwilson:
1.  Pass names of files to be built to `bin/knit_lessons.sh`.
2.  Only try to invoke `Rscript` if there are some files to knit.


fmichonneau:
that looks good to me

PR #65, opened 2016-06-24
move data/ into _episodes/ and _episodes_rmd/

fmichonneau:
`generate_md_episodes()` is modified such that datasets storred into
_episodes_rmd/data/ needed to knit the episodes, are copied into
_episodes/data/, and are made available at the root of the site.

### DIFF
--- a/bin/generate_md_episodes.R
+++ b/bin/generate_md_episodes.R
@@ -1,15 +1,41 @@
-if (require("knitr")) {
-    if (packageVersion("knitr") < '1.9.19') {
-        stop("knitr must be version 1.9.20 or higher")
-    }
-} else stop("knitr 1.9.20 or above is needed to build the lessons.")
+generate_md_episodes <- function() {
 
-if (!require("stringr"))
-    stop("The package stringr is required for generating the lessons.")
+    if (require("knitr")) {
+        if (packageVersion("knitr") < '1.9.19') {
+            stop("knitr must be version 1.9.20 or higher")
+        }
+    } else stop("knitr 1.9.20 or above is needed to build the lessons.")
 
-src_rmd <- list.files(pattern = "??-*.Rmd$", path = "_episodes_rmd", full.names = TRUE)
-dest_md <- file.path("_episodes", gsub("Rmd$", "md", basename(src_rmd)))
+    if (!require("stringr"))
+        stop("The package stringr is required for generating the lessons.")
 
-mapply(function(x, y) {
+    ## where the Rmd files and the datasets are located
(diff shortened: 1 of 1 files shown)


gvwilson:
I like the final result, but we're going to have to get there by:

1.  Creating a new `_episodes_rmd/data` directory.
2.  `git mv`'ing the data files there.

because the `sites` repository will still have `./data`, and if we relocate that in the downstream, we'll have merge headaches.  Also, do you want to copy `_episodes_rmd/data` into `_episodes/data` in `bin/generate_md_episodes.R`, or do you want the Makefile to take care of that with a separate command in the `lesson-rmd` target?

fmichonneau:
Looking at this with fresher eyes, I'm now wondering if it wouldn't be preferable to keep `./data` at the root, and copy the files temporarily to `./_episodes_rmd/data` for the knitting process (they will be deleted once the markdown files are generated). There will be a single copy of the data and we won't have to worry about merge conflicts.

The drawback I can see is that if maintainers or lesson contributors want to knit a single episode, they will need to use the Make file, otherwise the scripts will fail as there will be no `./_episodes_rmd/data`. What do you think?

gvwilson:
If the data files are in `data`, what path is used to open them from 
.Rmd files in `_episodes_rmd` during authoring? If possible, we'd like 
people to be able to edit .Rmd files interactively, testing things out 
as they go along - is that possible?

fmichonneau:
To make the data available during authoring, we could do the copying of the files in the temporary `./_episodes_rmd/data` at the beginning of each episode. This could be done in the `chunk-options.R` file that each episode has to evaluate. That would allow the knitting of individual files without having to use the Makefile, but doesn't allow interactive evaluation of the Rmd without editing the path of copying the data. This could be facilitated by having a commented out line of code, that could be evaluated if someone wants to work interactively with the Rmd file.

In summary, we need to decide whether we prefer to have 2 copies of the data in the repository (leading to possible confusion/diverging version), or if we make interactive editing a little difficult. I'm +0.5 for the latter.

gvwilson:
Hi François, does this *move* the ./data folder, or *add* a new _episodes_rmd/data folder?  I'd like to do the latter for now - having two data folders shouldn't be confusing, since people will either be writing .Rmd or not.

fmichonneau:
The data folder now lives at `_episodes_rmd/data`. (Its content isn't copied into `./_site` though so it won't be available through the published website).

fmichonneau:
it moves the `./_data` folder. I'll update the PR.

PR #67, opened 2016-06-25
Only cleaning intermediate R files on demand.

### LINKED ISSUE #66: Create separate Makefile target for deleting intermediate R files
The intermediate R files should not be deleted by `make clean`, since they have to be committed to the repository for GitHub Pages to work.

gvwilson:
1.  Splitting commands to clean intermediate R files into separate target.  These files must be committed to the repo, so they shouldn't be deleted under normal circumstances.
2.  Modifying intermediate file generation to create fig/rmd-* rather than fig/swc-rmd-*, since this template is used by other carpentries (not just SWC).

Closes #66.


fmichonneau:
It looks good to me. thanks!

PR #83, opened 2016-07-20
[fix # 82] Check R packages required by the lesson

fmichonneau:
Implements the solution proposed by @noamross to check that all the packages required to build the lessons are available, and if they are not they get installed.


gvwilson:
I trust you.

PR #89, opened 2016-07-31
Prevent wrapping of code fragments

gvwilson:
Taken from https://github.com/swcarpentry/lesson-template/pull/297 by @DanielBaird.


gvwilson:
Causes all sorts of unwanted weirdness.

PR #108, opened 2016-10-07
Starting to incorporate Library Carpentry materials

gvwilson:
Adding native support for Library Carpentry.


k8hertweck:
The general consensus from the recent maintainers meetings is that these changes won't impact our own lessons, but provide flexibility for folks wanting to develop Library Carpentry materials. However, the relationships between SWC/DC and LC are still [not clear](https://github.com/data-lessons/librarycarpentry/issues/6). Ultimately, it seems appropriate to encourage fledgling lessons, and worse case scenario, we remove references to LC later. @tracykteal, anything else?

tracykteal:
Agreed, this mainly just provides flexibility. +1

PR #109, opened 2016-10-07
Adding JQuery debugging

gvwilson:


rgaiacs:
@gvwilson This is still relevant? Can you provide a user case?

gvwilson:
We're still using JQuery, so having symbols when we debug may 
occasionally be useful.

Issue #115, opened 2016-12-14
2em padding on the left for lists


rgaiacs:
@maxim-belkin created https://github.com/swcarpentry/lesson-example/pull/83 for `lesson.scss` but I think that this maybe need to be fixed for other pages.

@gvwilson Could you point someone that had worked on the CSS more than me to help with this one?

maxim-belkin:
@rgaiacs thanks for pointing to the right repo. I can submit the same PR here if necessary.

Just to clarify, the issue with numbered lists can be seen [here](http://swcarpentry.github.io/lesson-example/).

@rgaiacs Also, I'm not entirely sure if you are asking @gvwilson to point to a person who worked on CSS for swcarpentry's repos or in general. In either case, the change that is being proposed is rather trivial.

rgaiacs:
I mean "a person who worked on CSS for swcarpentry's repos". After we "merge" all the CSS files I lost track of them. I want to avoid that this issue appear on other websites. :-)

PR #116, opened 2016-12-14
lesson.scss: 2em left padding for lists

maxim-belkin:
Example of the problem: see "Ten Things You Need To Know" from http://swcarpentry.github.io/lesson-example/

For numbered lists that have more that 9 elements 1 em padding on the left is not sufficient. The result is that "1" in numbers "10" and above is printed on top of the border of encapsulating HTML block. 2 em padding solves this problem for lists with less that 100 elements. It should be safe to assume that numbered lists in Carpentries' lessons should not have more than 99 items.

This PR is related to: #115 and swcarpentry/lesson-example#83

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -133,7 +133,7 @@ div.branding {
 
 ul,
 ol {
-  padding-left: 1em;
+  padding-left: 2em;
 }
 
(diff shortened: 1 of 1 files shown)


gvwilson:
@rgaiacs I think this is good to merge.

Issue #76, opened 2016-07-06
Formatting for in-line code should have background?


gdevenyi:
I think perhaps the in-line code segments would stand-out more if they had a background, like the classic github style.

Like `here`

gvwilson:
They would stand out more, but a lot of people find that distracting.  
Take it to a vote of the maintainers?

gdevenyi:
Yeah, reasonable.

fmichonneau:
+1

shwina:
-1 (somewhat reluctantly). I like in-line code background, but in portions with "dense" inline code, it can become highly distracting.

iglpdc:
+1 (but agreeing with @shwina on large portions of inline). I find the current version a bit hard to read, because you "feel" a difference, but sorta cannot say what's going on.

Maybe we could limit the use of inline code when very short (filenames, program and library names) and move the rest (any statement or command) to its own paragraph.

gdevenyi:
Agree, inline code is for short snippets, if large chunks exist that's a lesson formatting issue that should be fixed.

shwina:
What I meant was paragraphs like these:

Et `utinam theophrastus` has, `legimus` instructior pro `ut`. `Qui te` nobis prompta, `id` `quo` `quod quodsi` fierent. No fuisset `adolescens` sea, vocent `aeterno aperiam` id mel, quo no `labore` audire officiis. Vix et purto verterem vituperata, ex elit mentitum `urbanitas vis`.

Maybe it's just me that finds these disorientating :-)

gvwilson:
@k8hertweck can you please poll maintainers 'n' point them at this issue to see if they have opinions?

daattali:
+1 I came here to suggest this issue.

I do understand that sometimes it may look ugly, but it makes it much clearer what is code and what is text/punctuation. I just went over the entire unix lesson, and there were many times where I wasn't sure where in a paragraph the inline code ends, and if the punctuation is part of the command or not.

One example: [here](http://swcarpentry.github.io/shell-novice/04-pipefilter/)

> run the command `wc *.pdb`. wc is the

Without the background It's very easy to think that the period and second `wc` are also part of the command

rgaiacs:
Close by #117.

PR #117, opened 2016-12-14
Visually separate inline code blocks

maxim-belkin:
Change text color, background color and padding around inline code blocks.

### Before
before

### After
after

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -100,9 +100,9 @@ $codeblock-padding: 5px !default;
 //----------------------------------------
 
 code {
-  padding: 0 0;
-  color: inherit;
-  background-color: inherit;
+  padding: 2px 5px;
+  color: #3d90d9;
+  background-color: #e7e7e7;
 }
 
 img {


gvwilson:
I find this kind of highlighting more distracting than useful, particularly on smaller devices - @rgaiacs and @k8hertweck, shall we put it to a vote? And @maxim-belkin, thank you for the styling - if you have a few minutes to tackle the issue of vertical spacing after list items containing multiple paragraphs, we'd be very grateful.

rgaiacs:
> shall we put it to a vote?

I believe that this is one of the things that the new "Lesson Development" committee is going to vote on.

maxim-belkin:
I can look into
> the issue of vertical spacing after list items containing multiple paragraphs

if someone could point me to an example of it.

As for the PR: the purpose is to provide better visual separation of the code and the text. It can be broken into changes to text and/or background color, and font boldness/style. Currently employed method is `font style' which is clean but not easy to spot.

maxim-belkin:
related to #76

gvwilson:
Hard to show, because we've eliminated all instances from the existing 
lesson example :-)  To re-create, paste something like this into a page:

~~~
first term
:   first paragraph of definition

second term
:   second paragraph of definition
     *   bullet list
     *   in second definition

     with another paragraph in the definition

third term
:   more text
~~~

rgaiacs:
Tracy, Naupaka, Trever, Remi, and I approved this.

Issue #91, opened 2016-08-03
Only rebuild Markdown from R Markdown when necessary


gvwilson:
For some reason the Makefile always rebuilds all of the Markdown files corresponding to R Markdown files.  This shouldn't happen.

PR #119, opened 2016-12-31
Knit episodes only when any one RMarkdown source changes

itcarroll:
The recipe for any phony target (e.g. `lesson-rmd`) will always run, so that's why bin/knit_lessons.sh was always executed as described in issue #91.

This PR introduces the logic that the Markdown files in `RMD_DST` are the prerequisites for a single `lesson-md` target, which is just a dummy grouping variable. The prerequisite for any one of RMD_DST is the original array of RMarkdown files (i.e. RMD_SRC). On systems that run make in parallel, the line `${RMD_DST} : ${RMD_SRC}` could cause bin/knit_lessons.sh to run for each member of RMD_DST. That would be bad, hence `.NOTPARALLEL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ DST=_site
 
 # Controls
 .PHONY : commands clean files
+.NOTPARALLEL:
 all : commands
 
 ## commands         : show all commands.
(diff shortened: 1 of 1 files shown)


fmichonneau:
that looks good to me, but someone with more Make experience should also confirm.

rgaiacs:
@itcarroll Thanks for the pull request. I will have a look on it later.

gvwilson:
I just tested this one and it works as advertised - thanks @itcarroll, this is a great help. @rgaiacs @fmichonneau, I think it's ready to merge.

itcarroll:
@gvwilson - Sure thing. I sorted that out for [SESYNC's lesson master](https://github.com/sesync-ci/lesson-style), so I'm glad it's useful elsewhere.

PR #121, opened 2017-01-09
Using page.previous and page.next for links instead of loop

gvwilson:
This will make page builds faster and somewhat simpler to understand.


rgaiacs:
@gvwilson I'm merging this but I would like to know why we couldn't use it a few months ago?

gvwilson:
No idea - I remember trying it and not being able to get it to work, but 
that could easily have been operator error.

PR #153, opened 2017-06-23
Only convert rmd files to md if they have changed

mawds:
I have modified the build process so that only the R Markdown files in _episodes_rmd/ that have changed are converted to markdown when running  make lesson-md 

Previously all the episodes were rebuilt if any R markdown files were changed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,9 +83,8 @@ HTML_DST = \
 ## lesson-md        : convert Rmarkdown files to markdown
 lesson-md : ${RMD_DST}
 
-# Use of .NOTPARALLEL makes rule execute only once
-${RMD_DST} : ${RMD_SRC}
-	@bin/knit_lessons.sh ${RMD_SRC}
+_episodes/%.md: _episodes_rmd/%.Rmd
+	@bin/knit_lessons.sh $< $@
 
 ## lesson-check     : validate lesson Markdown.
 lesson-check :
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ all : commands
 commands :
 	@grep -h -E '^##' ${MAKEFILES} | sed -e 's/## //g'
 
+## serve-rmd        : run a local server, updating Rmd file automatically
+serve-rmd: lesson-md lesson-watchrmd serve
+
 ## serve            : run a local server.
 serve : lesson-md
 	${JEKYLL} serve
@@ -54,7 +57,7 @@ workshop-check :
 ## ----------------------------------------
 ## Commands specific to lesson websites.
 
-.PHONY : lesson-check lesson-md lesson-files lesson-fixme
+.PHONY : lesson-check lesson-md lesson-files lesson-fixme lesson-watchrmd
 
 # RMarkdown files
 RMD_SRC = $(wildcard _episodes_rmd/??-*.Rmd)
@@ -82,6 +85,9 @@ HTML_DST = \
 
 ## lesson-md        : convert Rmarkdown files to markdown
 lesson-md : ${RMD_DST}
+
+lesson-watchrmd:
+	@bin/watchRmd.sh &
 
 _episodes/%.md: _episodes_rmd/%.Rmd
 	@bin/knit_lessons.sh $< $@
--- a/bin/generate_md_episodes.R
+++ b/bin/generate_md_episodes.R
@@ -22,14 +22,17 @@ generate_md_episodes <- function() {
         install.packages(missing_pkgs)
     }
 
-    ## find all the Rmd files, and generate the paths for their respective outputs
-    src_rmd <- list.files(pattern = "??-*.Rmd$", path = "_episodes_rmd", full.names = TRUE)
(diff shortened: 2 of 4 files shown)


rgaiacs:
@mawds Thanks for the pull request.

> I have modified the build process so that only the R Markdown files in _episodes_rmd/ that have changed are converted to markdown when running make lesson-md
>
> Previously all the episodes were rebuilt if any R markdown files were changed

I don't know what behaviour writers want. I believe that we had `make lesson-md` building all the files to avoid problems with one file be compiled with version X and another file with version Y of R or a R package. I understand that it is annoying to wait minutes to have all the files compiled again. I wish that we were using GitLab Pages so that we could compile the Markdown files in a continuous integration environment but I will not raise that request again.

@swcarpentry/r-maintainers Could you review this to me? Thanks in advance.

mawds:
You can get a local continuous integration environment with the above commit; this watches the Rmd files using inotify-tools and compiles them on write.

jsta:
I tested the first part of this PR. It works for me and I prefer this setup (only building to md if the Rmd has changed). 

@mawds I am not familiar with `inotifywait`. Can you give us a recommended workflow to test it?

naupaka:
One issue may be that `inotifywait` is not installed by default on MacOS/Darwin, and looks not to be available on homebrew. [fswatch](https://github.com/emcrisostomo/fswatch) seems like the closest analogue.

jsta:
Is there a way to have two separate commands for each like `make lesson-md` and `make lesson-md-all` or `make lesson-md-force`?

mawds:
@naupaka I wasn't aware of fswatch; it looks like this might be a better option.  I'll talk a look at using it instead.

rgaiacs:
Please see https://github.com/swcarpentry/styles/issues/171.

mawds:
I've modified the code to use fswatch or inotifywait.    Unfortunately fswatch picks up multiple "update" events when you write a file in RStudio (so the .Rmd to .md conversion happens multiple times).  

I've also realised that inotifywait doesn't detect changes when using vim; fswatch does (and only picks up one change).   

So, in summary, it looks like fswatch is the better approach if the detection of multiple updates in a short time window could be suppressed.   I'm not sure how best to do this. 

(@rgaiacs - I don't think #171 does quite what I'm trying to achieve;  I'd like the .md file to be generated automatically once the .Rmd file is saved.  My workflow is to have the episode page open in Firefox, with the [auto reload](https://addons.mozilla.org/en-GB/firefox/addon/auto-reload/?src=api) plugin watching the page, so I get to see updates in (near) real time)

rgaiacs:
@mawds I'm happy with any decision from @swcarpentry/r-maintainers.

mawds:
Thanks @rgaiacs ; my feeling is that unless we can deal with the issue of multiple updates for a single save we _shouldn't_ use the fswatch/inotifywait approach.   I'll give some thought to how we might do this, but any suggestions would be very welcome.

rgaiacs:
> my feeling is that unless we can deal with the issue of multiple updates for a single save we shouldn't use the fswatch/inotifywait approach.

You can have a look at https://github.com/rstudio/blogdown/ and see what they are using. From my look, see https://github.com/rstudio/blogdown/blob/master/R/serve.R, they are using [servr](https://cran.r-project.org/web/packages/servr/index.html)::[watch_dir](https://github.com/rstudio/blogdown/blob/master/R/serve.R#L108).

rgaiacs:
Sorry, I don't know what trigger this pull request to be close.

rgaiacs:
@fmichonneau I'm assigning this pull request to you because my R skills are not amazing. :-)

mawds:
I wasn't able to work out  a robust, cross platform and cross editor way of working out when a file had changed (and only picking up that change once).  Unless anyone else got any further I'd suggest closing this.

maxim-belkin:
I don't understand the purpose of this PR because `make` exists specifically for this purpose: execute a set of commands when dependencies are newer than the target.

I don't have much time until this Saturday, so let me ping @gcapes

gcapes:
I think the purpose of the PR is to auto-rebuild a .md file when the corresponding .Rmd file is saved, so that you can view your edits locally in real time using Jekyll, as you would for any other lesson.
I think the Makefile works just fine, but the idea was to avoid having to re-run Make after every single edit.

However, @mawds hasn't found a robust solution no-one else has commented since, so maybe best to close this PR?

rgaiacs:
> I think the purpose of the PR is to auto-rebuild a .md file when the corresponding .Rmd file is saved

True.

> so maybe best to close this PR?

Some times I prefer to keep the issue pull request open to reminder our future self of what we could do to improve our workflow.

jsta:
My preferred behavior is at https://github.com/datacarpentry/r-raster-vector-geospatial/commit/2a3bf5e4382a6c9dc57df66e78dc0e79ddce8ff8#diff-b67911656ef5d18c4ae36cb6741b7965. A full rebuild could be called with `make -B serve`.

fmichonneau:
replaced by #321

PR #163, opened 2017-08-08
Output episode figures in correct order

mawds:
make lesson-figures was generating figures out of episode order.  This appears to be due to a change in GNU Make. In older versions this returned values from $(wildcard) in sorted order; it no longer does so.  See https://stackoverflow.com/questions/40558385/gnu-make-wildcard-no-longer-gives-sorted-output-is-there-any-control-switch

This PR sorts the output of $(wildcard) so that episode figures appear in the correct order.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,9 +65,9 @@ MARKDOWN_SRC = \
   index.md \
   CONDUCT.md \
   setup.md \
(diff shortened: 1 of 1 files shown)


rgaiacs:
@mawds Thanks for the pull request.

Issue #100, opened 2016-09-02
Style for keyboard keys


rgaiacs:
@datacarpentry has some CSS for keyboard keys:

![screenshot_2016-09-02_13-50-45](https://cloud.githubusercontent.com/assets/1506457/18204394/7828b9ec-7114-11e6-8262-fd14154910ef.png)

Can we have it?

gvwilson:
How do I express this in Markdown without using HTML tags?

rgaiacs:
~~~
$ kramdown <<EOF
`TAB`{: .keyboard }
EOF
<p><code class="keyboard">TAB</code></p>
~~~

, as describe at http://kramdown.gettalong.org/quickref.html#inline-attributes, and

~~~
.keyboard {
    padding: 2px 4px;
    font-size: 90%;
    color: #fff;
    background-color: #333;
    border-radius: 3px;
    -webkit-box-shadow: inset 0 -1px 0 rgba(0,0,0,.25);
    box-shadow: inset 0 -1px 0 rgba(0,0,0,.25);
    font-family: monospace;
}

gvwilson:
Ick.

brandoncurtis:
There has been some confusion about keyboard key name formatting, in e.g. swcarpentry/shell-novice/pull/561

Regarding the proposal above:

<pre>when the user types a command and then presses `ENTER`{: .keyboard } ...</pre>

That doesn't look so bad to me!  @gvwilson, is that definitely a no-go for you?

We have this [in StackExchange in the form of an HTML tag][sx], though I gather that HTML tags should be avoided in SWC material.

The alternative would be to add something to the SWC styleguide regarding formatting of keyboard key names, such as "all caps and surrounded by backticks":

<pre>`ENTER`</pre>

(Are the contents of https://github.com/swcarpentry/lesson-example the definitive styleguide?)

[sx]: https://meta.stackexchange.com/questions/26207/how-can-i-format-as-keyboard-keys

brandoncurtis:
I explored this a little further and *ENTER*{: .keyboard } is preferred over `ENTER`{: .keyboard} so that Unicode string references work for e.g. the arrow keys.

It looks like this:
![keys-demo](https://user-images.githubusercontent.com/2591290/29035841-e73333ec-7b51-11e7-9772-6b2a44b12127.png)


With this CSS:

```css
// keyboard key name style, from StackExchange
.keyboard {
    display: inline-block;
    margin: 0 .1em;
[...]

PR #165, opened 2017-08-18
add keyboard key style

brandoncurtis:
Proposed usage:

<pre>when the user types a command and then presses *Enter*{:.keyboard} ...</pre>

The CSS is borrowed from StackExchange, which [implements this with the `<kbd>` tag][sx].

Block inline attribute lists are already used by SWC for e.g. the codeblocks, so I gather that this [span inline attribute list][sial] implementation is preferred over an HTML tag for SWC material.

SWC prefers verbose identifiers (`.objectives`, `.keypoints`, &etc), so I think `{:.keyboard}` would be preferred over `{:.kbd}`.

In practice it looks like this:

keyboard-key-style

This would settle some confusion about how to format keyboard key names, such as swcarpentry/shell-novice#561.

[sx]: https://meta.stackexchange.com/questions/26207/how-can-i-format-as-keyboard-keys
[sial]: https://kramdown.gettalong.org/syntax.html#span-ials

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -153,3 +153,25 @@ span.fold-unfold {
   margin-left: 1em;
   opacity: 0.5;
 }
+
+
+//----------------------------------------
+// keyboard key style, from StackExchange.
+//----------------------------------------
+
+.keyboard {
+    display: inline-block;
+    margin: 0 .1em;
+    padding: .1em .6em;
+    font-family: Arial,"Helvetica Neue",Helvetica,sans-serif;
+    font-size: 11px;
+    line-height: 1.4;
+    color: #242729;
+    text-shadow: 0 1px 0 #FFF;
+    background-color: #e1e3e5;
+    border: 1px solid #adb3b9;
+    border-radius: 3px;
+    box-shadow: 0 1px 0 rgba(12,13,14,0.2), 0 0 0 2px #FFF inset;
+    white-space: nowrap;
+    font-style: normal;
+}
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -159,7 +159,7 @@ span.fold-unfold {
 // keyboard key style, from StackExchange.
 //----------------------------------------
 
-.keyboard {
+kbd {
     display: inline-block;
     margin: 0 .1em;
     padding: .1em .6em;


brandoncurtis:
originally proposed in #100

rgaiacs:
@brandoncurtis Thanks for the pull request. Please see [my message](http://lists.software-carpentry.org/pipermail/maintainers/2017-August/000370.html) to the maintainers list.

# Comments

I believe that we agreed in the past that accessibility is one of our "core principles". We have a [accessibility checklist](https://software-carpentry.org/checklists/accessibility/), a [blog post](https://software-carpentry.org/blog/2016/09/teaching-programming-to-the-blind.html) about teaching visually impaired learners and [some discussion](http://lists.software-carpentry.org/pipermail/discuss/2016-September/004737.html) on the mailing list. Because of that I will strong prefer to use HTML tags instead of a CSS workaround, i.e. use

~~~
when the user types a command and then presses <kbd>Enter</kbd> ...
~~~

instead of

~~~
when the user types a command and then presses *Enter*{:.keyboard} ...
~~~

**Note**: [This answer in StackOverflow](https://stackoverflow.com/a/15190727/1802726) says

> There does not seem to be any evidence of any particular handling of the `kbd` element in browsers, assistive software, or otherwise, except for the default rendering (which uses the browser’s default monospace).

but this is mostly a implementation problem. If we are trying to add semantic to the text we should to it following the specification.

cmacdonell:
I'd say I prefer the look of @brandoncurtis suggestion. I understand how <kbd> fits with our accessibility goal, but if browsers and/or screen-readers don't handle <kbd> then I would say to go with CSS workaround.

iglpdc:
I think we should definitely add the style, but I'm not sure which is the best way.

naupaka:
I am a little confused. Why won't the `<kbd>Enter</kbd>` solution and related CSS do what we want?

brandoncurtis:
@rgaiacs, @naupaka, @iglpdc, @cmacdonell I meditated on this for a bit and I prefer <kbd>.  I've updated this PR.  Additional thoughts?

naupaka:
Looks good to me!

brandoncurtis:
@naupaka correct!  html tag and CSS styles.  I'm also planning to add a usage note to the example lesson in https://github.com/swcarpentry/lesson-example.

rgaiacs:
@brandoncurtis Thanks for the work. Could you send a pull request to  https://github.com/swcarpentry/lesson-example covering how people should use this? Some things that I would like that you clarify

1. Should we use `<kbd>X</bbd>` or `<kbd>X</bbd>` for lowercase letters? 
2. Should we use `<kbd>Shift</bbd>+<kbd>X</bbd>` or `<kbd>Shift</bbd> + <kbd>X</bbd>`?

screencapture-file-tmp-space-html-1504000494034

3. What unicode writers should use for arrow?

~~~
<p>Many arrows:</p>
<ul>
  <li><kbd>←</bbd></li>
  <li><kbd>↤</bbd></li>
  <li><kbd>⟵</bbd></li>
  <li><kbd>⟻</bbd></li>
</ul>
~~~

screencapture-file-tmp-space-html-1504000980551

Thanks

rgaiacs:
@brandoncurtis Thanks for all the help and patience. I talked with other people and we have official green light to merge this after we add some documentation as mentioned on the previous comment.

rgaiacs:
@brandoncurtis I submited https://github.com/swcarpentry/lesson-example/pull/131 which address some of my comments at https://github.com/swcarpentry/styles/pull/165#issuecomment-325617688. Your comments are very welcome. Thanks

vuw-ecs-kevin:
As this issue has just passed through my mail box, so this is probably
a bit late now but

  A cautionary tail about using images, where words impart information.

An engineer and I once spent way too long in the machine room trying
to get to the "monitor level" prompt of some hardware, (this often requires
pressing keys quickly before the normal handler kicks in) despite following
the instructions that had been sent to us, at short notice.

What the instructions were missing were the two lovely pictures of the 
"Space" key that had appeared in the original on-line manual but which 
had not been rendered in the version sent to us. Had those lovely pictures
been combined with the plain text they were taking the place of, we'd have
been a lot warmer when we exited the machine room.

Takeaways:

1. The text in an ALT-tag should leave should less room for confusion 
2. People will print things out, from within a browser, so make sure the
    pictures don't end up hiding the real message.

Just my thr'pen'th.

katrinleinweber:
Hi! What about the Mac's `cmd` key? That's how it is physically etched-in, but should we spell it out as `Command` or use `Cmd`?

rgaiacs:
@katrinleinweber Thanks for the question. Please see https://github.com/carpentries/lesson-example/pull/228.

PR #166, opened 2017-08-28
Allows for fully-offline-capable functionality in lesson navigation.

vuw-ecs-kevin:
Note
---
This patch is also the *alternative approach* mentioned in Raniere's [PR 127](https://github.com/swcarpentry/lesson-example/pull/127).

Description
---
Links in the lesson are of the form `http://swcarpentry.github.io/LESSON-NAME/XX-EPISODE-NAME/`. The links works when the lesson is being server by a web server because the server know that it should return the file `_site/XX-EPISODE-NAME/index.html` but will not work if viewing the lesson over file URI scheme because `file:///PATH/TO/LESSON/ROOT/_site/XX-EPISODE-NAME/` is not the same as `file:///PATH/TO/LESSON/ROOT/_site/XX-EPISODE-NAME/index.html`.

Proposal
---
Allow for rendering of internal episode content as 

`scheme:/path/to/LESSON-NAME/XX-EPISODE-NAME/index.html` 
   instead of 
`scheme:/path/to/LESSON-NAME/XX-EPISODE-NAME/`

and of internal auxilliary content as, for example,

`scheme:/path/to/LESSON-NAME/discuss/index.html` 
  instead of 
`scheme:/path/to/LESSON-NAME/XX-EPISODE-NAME/discuss`

Benefits
---
As mentioned before, ability to correctly navigate the lesson without running a web server.

Drawback
---
None: if the proposed variable is left blank.

If the proposed variable is set to `index.html`, link targets are 10  characters longer.

Implementation
---
Applies a `{{ site.index }}` variable which, by default, is **blank**, in which
case all links render as though the variable had not been introduced, 
but which, when set to `index.html`, produces the same link URIs that 
a typical web server would create on the fly.

Alternative Implementation
--- 
Use the variable name `{{ site.permalink }}` as that may mean more to people
who think of the filename in a URI as "permalinks", as in, from a *web page* 
standpoint, as opposed the default index, as in, from a *web server* standpoint,
and which matches the existing nomenclature with the styles template.

Related work
---
This patch is effectively just [vuw-ecs-kevin:offline-capable-lesson@66442fc](https://github.com/vuw-ecs-kevin/offline-capable-lesson/commit/66442fcdf40f967302bc714c4b693f9355eb4061) applied into the swcarpentry:styles repository, the only difference being that the change to  `_config.yml`  is made to  `bin/lesson_initialize.py`  which creates  `_config.yml`.

See https://github.com/vuw-ecs-kevin/offline-capable-lesson/ for the background.

### DIFF
--- a/_includes/all_keypoints.html
+++ b/_includes/all_keypoints.html
@@ -7,7 +7,7 @@
   {% unless episode.break %}
     <tr>
       <td class="col-md-3">
-        <a href="{{ page.root }}{{ episode.url }}">{{ episode.title }}</a>
+        <a href="{{ page.root }}{{ episode.url }}{{ site.index }}">{{ episode.title }}</a>
       </td>
       <td class="col-md-9">
         <ul>
--- a/_includes/episode_navbar.html
+++ b/_includes/episode_navbar.html
@@ -5,23 +5,23 @@
   <div class="col-md-1">
     <h3>
       {% if page.previous.url %}
-      <a href="{{ page.root }}{{ page.previous.url }}"><span class="glyphicon glyphicon-menu-left" aria-hidden="true"></span><span class="sr-only">previous episode</span></a>
+      <a href="{{ page.root }}{{ page.previous.url }}{{ site.index }}"><span class="glyphicon glyphicon-menu-left" aria-hidden="true"></span><span class="sr-only">previous episode</span></a>
       {% else %}
-      <a href="{{ page.root }}/"><span class="glyphicon glyphicon-menu-up" aria-hidden="true"></span><span class="sr-only">lesson home</span></a>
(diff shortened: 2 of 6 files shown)


rgaiacs:
> his patch is effectively just vuw-ecs-kevin:offline-capable-lesson@66442fc applied into the swcarpentry:styles repository, the only difference being that the change to _config.yml is made to bin/lesson_initialize.py which creates _config.yml.

I regret our choice to use `bin/lesson_initialize.py` to create some files because it just make propagate the changes much harder. :-(

rgaiacs:
@vuw-ecs-kevin Thanks for your time to put this pull request in place. As I mentioned in https://github.com/swcarpentry/lesson-example/pull/127, I will go with the solution that I put in place because it will reduce the code to be maintain. I'm closing this pull request for now as I will need to do some Git magit to merge my pull request.

PR #169, opened 2017-09-05
Load methods package

jsta:
Building `md` from `Rmd` in a script like this will break if the `methods` package is not called and the code includes class object routines. See https://github.com/r-spatial/sf/issues/478


fmichonneau:
+1

rgaiacs:
@jsta Thanks very much for the pull request. Looks good to me.

PR #170, opened 2017-09-08
Adds capability to render MultiDay Lessons as MultiPart Lessons

vuw-ecs-kevin:
No change to default rendering of lesson content
    Requires extra site variable "parts" in conjunction with "start"
     to alter rendering.

---

SWC's style template currently allows for MultiDay lessons
to be displayed in the Schedule as follows:

```
 		Setup 			Download files ...
Day 1 	09:00 	1. Introduction 	Key question
	09:30 	2. Second episode 	Key question
	10:00 	3. Third episode 	Key question
	10:30 	Finish 	
Day 2 	09:00 	4. Fourth episode 	Key question
	09:30 	5. Fifth episode 	Key question
	10:00 	Finish
```

In a recent email conversation amongst the Data Carpentry
Cloud Genomics maintainers, I had sugggested that one way
that we might seek to handle the situation whereby our
proto-Lesson was being derived from a number of seperate
existing resources (a number of which duplicated the core
episode themes) within the overall Lesson syllabus being
proposed, would be to make the `gh-pages` branch a MultiDay
Lesson, so that the episodes from the separate resources
could all be displayed, in order, but be separated into
multiple days, until such time as we might refactor the
content.

Thinking that the "Days' approach seemed a little clumsy,
I also had the idea that a temporary change to

`_includes/syllabus.html`

whereby the word "Day" would be replaced by "Part", would
look a little less clumsy, but then started thinking about making 
that change less of a temporary one, but something that
could be a user-effected choice.

This PR takes the idea that some SWC Lessons might benefit
from being laid out in seperate Parts and so adds to the
current styles template, via a small set of changes to

`_includes/syllabus.html`

the capability to make a multiday lesson be rendered as

```
 		Setup 			Download files ...
Part 1	09:00 	1. Introduction 	Key question
	09:00 	2. Second episode 	Key question
	10:00 	3. Third episode 	Key question
	10:30 	Finish 	
Part 2	10:30 	4. Fourth episode 	Key question
	11:00 	5. Fifth episode 	Key question
	11:30 	Finish
```

Note that the the Mutliday's resetting of the start time for each
new day is removed when breaking up a Lesson into Parts.

That does beg the question as to how to handle MultiPart Lessons
held within Multiday Lessons, but that's one for the future!

---
Kevin M. Buckley

eScience Consultant
School of Engineering and Computer Science
Victoria University of Wellington
New Zealand

### DIFF
--- a/_includes/syllabus.html
+++ b/_includes/syllabus.html
@@ -1,6 +1,7 @@
 {% comment %}
   Display syllabus in tabular form.
   Days are displayed if at least one episode has 'start = true'.
+  Days are displayed as Parts if at least one episode has 'part = true'.
 {% endcomment %}
 <div class="syllabus">
   <h2 id="schedule">Schedule</h2>
@@ -8,8 +9,12 @@
   {% assign lesson_number = 0 %}
   {% assign day = 0 %}
   {% assign multiday = false %}
+  {% assign partnotday = false %}
   {% for episode in site.episodes %}
     {% if episode.start %}{% assign multiday = true %}{% break %}{% endif %}
+  {% endfor %}
+  {% for episode in site.episodes %}
+    {% if episode.parts %}{% assign partnotday = true %}{% break %}{% endif %}
   {% endfor %}
   {% assign current = site.start_time %}
 
@@ -33,12 +38,14 @@
           <td class="col-md-7"></td>
         </tr>
       {% endif %}
-      {% assign current = site.start_time %} {% comment %}Re-set start time of this episode to general daily start time {% endcomment %}
+      {% if partnotday %}{% else %}
+        {% assign current = site.start_time %} {% comment %}Re-set start time of this episode to general daily start time {% endcomment %}
+      {% endif %}
     {% endif %}
     {% assign hours = current | divided_by: 60 %}
     {% assign minutes = current | modulo: 60 %}
     <tr>
-      {% if multiday %}<td class="col-md-1">{% if episode.start %}Day {{ day }}{% endif %}</td>{% endif %}
+      {% if multiday %}<td class="col-md-1">{% if episode.start %}{% if partnotday %}Part{% else %}Day{% endif %} {{ day }}{% endif %}</td>{% endif %}
       <td class="{% if multiday %}col-md-1{% else %}col-md-2{% endif %}">{% if hours < 10 %}0{% endif %}{{ hours }}:{% if minutes < 10 %}0{% endif %}{{ minutes }}</td>
       <td class="col-md-3">
         {% assign lesson_number = lesson_number | plus: 1 %}


_includes/syllabus.html
```diff
@@ -1,16 +1,21 @@
...
   <h2 id="schedule">Schedule</h2>
 
   {% assign lesson_number = 0 %}
   {% assign day = 0 %}
   {% assign multiday = false %}
+  {% assign partnotday = false %}
   {% for episode in site.episodes %}
     {% if episode.start %}{% assign multiday = true %}{% break %}{% endif %}
   {% endfor %}
+  {% for episode in site.episodes %}
```
rgaiacs: You can avoid this look by including the next line in the previous loop. This should make the compilation faster.

_includes/syllabus.html
```diff
@@ -33,12 +38,14 @@ <h2 id="schedule">Schedule</h2>
           <td class="col-md-7"></td>
         </tr>
       {% endif %}
-      {% assign current = site.start_time %} {% comment %}Re-set start time of this episode to general daily start time {% endcomment %}
+      {% if partnotday %}{% else %}
```
rgaiacs: Would this break the multi day?

_includes/syllabus.html
```diff
@@ -33,12 +38,14 @@ <h2 id="schedule">Schedule</h2>
...
+        {% assign current = site.start_time %} {% comment %}Re-set start time of this episode to general daily start time {% endcomment %}
+      {% endif %}
     {% endif %}
     {% assign hours = current | divided_by: 60 %}
     {% assign minutes = current | modulo: 60 %}
     <tr>
-      {% if multiday %}<td class="col-md-1">{% if episode.start %}Day {{ day }}{% endif %}</td>{% endif %}
+      {% if multiday %}<td class="col-md-1">{% if episode.start %}{% if partnotday %}Part{% else %}Day{% endif %} {{ day }}{% endif %}</td>{% endif %}
```
rgaiacs: The condition here looks wrong. Would you not need

~~~
{% if multiday %}...{% endif %}
{% if partnotday %}...{% endif %}
~~~

rgaiacs:
@vuw-ecs-kevin Thanks for the pull request and all yours last contributions. I think that this change need to be futher discussed with @tracykteal, @ChristinaLK, @k8hertweck et al.

vuw-ecs-kevin:
On Mon, September 11, 2017 20:43, Raniere Silva wrote:

> You can avoid this look by including the next line in the previous loop.
> This should make the compilation faster.

Yeah.

When I first scoped out the change, I had both checks inside
the one loop but then realised that the break in the "start"
check was causing the "parts" not to be seen.

I thus chose to use a second loop for clarity, rather than
change too much in the original loop.

I feel that if a proposed change looks "small" then it has more
chance of passing the first hurdle to being accepted.


> Would this break the multi day?

Not in the testing I have done so far, but I'm sure there are
"edge cases" catered for in the current styles template that
may not be apparaent to folk like me.

As with the simplicity of the "second loop" above is more of
a "proof of concept" than a "sine qua non" implementation of
that concept.


> The condition here looks wrong. Would you not need
>
> ~~~
> {% if multiday %}...{% endif %}
> {% if partnotday %}...{% endif %}
> ~~~

Not as I think of it, no.

The varibale "multiday" is the current vehicle by which a Lesson
is split into multiple parts.

The "partnotday" merely determines which of two possible labels
will get placed onto those parts.


Given that you have raied the question though, I could see a case,
assuming that a Lesson split into Parts, not just Days, is seen to
be desirable, for renaming "multiday" to be "multipart", at which
point "partnotday" might become "daysorparts" which more closely
reflects what it's actually controlling in the rendering.

Kevin

rgaiacs:
@vuw-ecs-kevin thanks for the delay.

> In a recent email conversation amongst the Data Carpentry
> Cloud Genomics maintainers, I had sugggested that one way
> that we might seek to handle the situation whereby our
> proto-Lesson was being derived from a number of seperate
> existing resources (a number of which duplicated the core
> episode themes) within the overall Lesson syllabus being
> proposed, would be to make the gh-pages branch a MultiDay
> Lesson, so that the episodes from the separate resources
> could all be displayed, in order, but be separated into
> multiple days, until such time as we might refactor the
> content.
>
> Thinking that the "Days' approach seemed a little clumsy,
> I also had the idea that a temporary change to
>
> _includes/syllabus.html
>
> whereby the word "Day" would be replaced by "Part", would
> look a little less clumsy, but then started thinking about making
> that change less of a temporary one, but something that
> could be a user-effected choice.

After think about it, I would prefer to drop the "Day" division in favour of the "Part" division and support more than one "Part", for example, morning of the first day and afternoon of the first day. Since only the Data Carpentry Cloud Genomics maintainers is proto-typing this division and Instructor Training is more than one day long, I'm trusting @ErinBecker and @tracykteal to select among

1. We keep the "Day" division and push the "Part" division to the future
2. We drop the "Day" division in favour of the "Part" division
3. There is a **real case** for "Day" division and "Part" division **instead of authors preference** and we need to support both

I'm working to freeze the styles on Friday, 27 October 2017 for the end of the year release. Please let me know of the decision so I could accommodate this on the next release.

rgaiacs:
Sorry, I don't know what trigger this pull request to be close.

_includes/syllabus.html
```diff
@@ -1,16 +1,21 @@
...
 
   {% assign lesson_number = 0 %}
   {% assign day = 0 %}
   {% assign multiday = false %}
+  {% assign partnotday = false %}
   {% for episode in site.episodes %}
     {% if episode.start %}{% assign multiday = true %}{% break %}{% endif %}
   {% endfor %}
+  {% for episode in site.episodes %}
+    {% if episode.parts %}{% assign partnotday = true %}{% break %}{% endif %}
```
maxim-belkin: According to [line 4](https://github.com/carpentries/styles/pull/170/files/fd66850de72315244eae8adc8b368d1335a08b2a#diff-6be1ef7cb9e3d5c3bd34049c20599594R4), it should be `episode.part`

_includes/syllabus.html
```diff
@@ -1,16 +1,21 @@
...
   Days are displayed if at least one episode has 'start = true'.
+  Days are displayed as Parts if at least one episode has 'part = true'.
 {% endcomment %}
 <div class="syllabus">
   <h2 id="schedule">Schedule</h2>
 
   {% assign lesson_number = 0 %}
   {% assign day = 0 %}
   {% assign multiday = false %}
+  {% assign partnotday = false %}
```
maxim-belkin: I'd suggest changing `partnotday` to `multipart`

_includes/syllabus.html
```diff
@@ -1,16 +1,21 @@
 {% comment %}
   Display syllabus in tabular form.
   Days are displayed if at least one episode has 'start = true'.
+  Days are displayed as Parts if at least one episode has 'part = true'.
```
maxim-belkin: I'd suggest renaming `start` to `new_day` and `part` to `new_part`

_includes/syllabus.html
maxim-belkin: Time must be reset here because this section is for a new day (see [line 29 above](https://github.com/carpentries/styles/pull/170/files#diff-6be1ef7cb9e3d5c3bd34049c20599594R29)).

_includes/syllabus.html
maxim-belkin: I think it should as follows: 
```
{% if multiday %}<td class="col-md-1">{% if episode.start %}Day {{ day }}{% endif %}</td>{% endif %}
{% if multipart %}<td class="col-md-1">{% if episode.part %}Part {{ part }}{% endif %}</td>{% endif %}
```

Note, I used `multipart` rather than `partnotday` and non-existent (at the moment) `part` variable

fmichonneau:
Thanks for this @vuw-ecs-kevin.

I think that the code complexity to add this feature does not balance out its benefits. If instructors want to update the schedule, they can edit it manually to fit their needs.

I'm going to close this PR given the lack of activity but if someone wants to make a case for it and update it, the lesson infrastructure committee could reconsider it.

PR #174, opened 2017-10-11
Fix episode nav wrapping

IanLee1521:
By doing this, when a page is heavily zoomed in (e.g. 500%) the navbar icons stay in the same row, rather than breaking onto two different rows. Also pulls the arrows to the outside of the row.

See also: https://github.com/carpentries/instructor-training/pull/521


rgaiacs:
Looks great. Thanks for your contribution.

Issue #198, opened 2018-01-11
Add comment to all files stating the origin


rgaiacs:
Is getting more common for contributors to do changes in files originated in this repository but send the pull request to another place, for example, https://github.com/swcarpentry/workshop-template/pull/458. Contributors are not to be blame!!! They are doing it on their best intentions. We need to add at the top of every file stating that the changes need to be send to this repository.

PR #200, opened 2018-01-25
Advise users not to edit .md files generated from .Rmd files

mawds:
This PR partially deals with #198, by adding a comment to the `.md` files that are generated from `.Rmd` files (see also swcarpentry/r-novice-gapminder/pull/340). This advises people not to edit the generated `.md` file, but instead to edit the `.Rmd` file.


rgaiacs:
@mawds Thanks. This looks good to me. Any @swcarpentry/r-maintainers can have a look at this and give me thumbs up to merge? Thanks in advance.

naupaka:
This looks very nice. Couple things I might change:

1. switch out the `sapply()` for a `vapply()` and, since the main purpose of this function is its side effect, have the function return a message indicating what happened. I've got a fear of silent `sapply()` failures ever since reading this chapter in [Advanced R](http://adv-r.had.co.nz/Functionals.html). This way if something fails somewhere, it'll be easier to troubleshoot, and if `dest_md` ever ends up not what we expect, we'll hear about it because the returned vector won't be of length 1. 
2. switch the `=` to `<-` on the second line for consistency and to follow the style guide.
3. `close()` the `con` when we're done with it.

Here's my slightly revised version:

```r
  # Read the generated md files and add comments advising not to edit them
  vapply(dest_md, function(y) {
    con <- file(y)
    mdfile <- readLines(con)
    if (mdfile[1] != "---")
      stop("Input file does not have a valid header")
    mdfile <- append(mdfile, "# Please do not edit this file directly; it is auto generated.", after = 1)
    mdfile <- append(mdfile, paste("# Instead, please edit", 
                                   basename(y), "in _episodes_rmd/"), after = 2)
    writeLines(mdfile, con)
    close(con)
    return(paste("Warning added to YAML header of", y))
  },
  character(1))
```

mawds:
Thanks @naupaka ; that's a better way of doing it.  I'll update the PR with your changes.

rgaiacs:
@mawds Thanks. I'm merging it now.

PR #203, opened 2018-02-12
fix svg files

fmichonneau:
@rgaiacs I'll let you merge this PR.


rgaiacs:
@fmichonneau Can you provide the steps that you follow to make the changes? I know that SVG is a plain text XML file but I can't read it as plain text and create a image in my brain so I would like to know what the changes are.

For example, I don't think that we need

~~~
xmlns:ooo="http://xml.openoffice.org/svg/export"
~~~

because the logo is not design to be edit in LibreOffice.

fmichonneau:
As mentioned [here](https://github.com/swcarpentry/lesson-example/issues/171#issuecomment-365330839) the files were rewritten using https://github.com/fmichonneau/fixsvg

As far as I can the only modification made to the content of the file has to do with the removal of the `clip-path` element and the reordering of the elements within the tags because of the way `libxml2` parses and writes XML files.

rgaiacs:
> As mentioned here the files were rewritten using https://github.com/fmichonneau/fixsvg

Thanks.

@fmichonneau Can I ask you to change https://github.com/swcarpentry/styles/blob/gh-pages/Makefile to

1. Include a rule to convert the `.odg` files to `.svg` files
2. Clean **all** the `.svg` files to using [your script](https://github.com/fmichonneau/fixsvg)

   I can hear instructors complaining about add more dependencies to our pipeline but this is a bigger issue that will take more time to resolve.
3. Review Makefile dependence tree.

When we have a pull request in place we can plan a new release to include this changes. Thanks in advance.

rgaiacs:
Firefox 58.0.2

Left: Pull request
Right: 34b0d399d6318028ca0496d06f2fa792d24837e5

screenshot from 2018-02-19 16-16-12

PR #216, opened 2018-03-18
Harmonise code style through .editorconfig

katrinleinweber:
Draft implementation of https://github.com/swcarpentry/styles/issues/131 's suggestion.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,10 +8,15 @@ trim_trailing_whitespace = true
 [*.md]
 indent_size = 2
 indent_style = space
+max_line_length = 100
+
+[*.r]
+max_line_length = 80
 
 [*.py]
 indent_size = 4
 indent_style = space
+max_line_length = 79
 
 [*.sh]
 end_of_line = lf
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.py]
+indent_size = 4
+indent_style = space
+
+[*.sh]
+end_of_line = lf
+
+[Makefile]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,10 @@ charset = utf-8
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.md]
+indent_size = 2
+indent_style = space
+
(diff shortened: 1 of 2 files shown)


rgaiacs:
@katrinleinweber Thanks for the pull request. To accept it I would like to have come continuous integration (see http://pydocs.editorconfig.org/en/latest/command_line_usage.html) to enforce contributions to follow the rules.

cc @fmichonneau

katrinleinweber:
Hey, come on ;-D Currently, nothing is enforced automatically. So this step to `.editorconfig` already provides some of that "enforcement". At the source. Subliminally. And largely without anyone having to lift a finger. A clear benefit is within reach.

Setting up a CI pipeline is barely within my reach and thus heavy lifting. I did want to learn it at some point anyway, but I can't promise how soon I'll find the time.

fmichonneau:
Thanks for this @katrinleinweber I'm going to look into this more. I'm not very familiar with using .editorconfig at the moment. Re: implementing the CI, you don't have to worry about this, it's something I'm  working on at the moment.

maxim-belkin:
@katrinleinweber, how about `[*.md]` section and `max_line_length` setting? Otherwise, this PR looks good to me.

Regarding CI: IMHO we should not bother with adding EditorConfig checks to CI...

katrinleinweber:
And which number of characters for `max_line_length`? 80 is RStudio's default, but [Python demands 79](https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

maxim-belkin:
in `lesson-check` it is set to 100... 🙄

katrinleinweber:
Ah, right! Thanks for the reminder. So, should we duplicate that piece of info here?

IMHO, it seems reasonable to let `.editorconfig` serve as the "1st line of defense" (so to speak) and "[lesson_check.py](https://github.com/swcarpentry/styles/blob/0e0f2bf4e396419a46760d0add6af054cd1c512c/bin/lesson_check.py#L102)" as the last. But puts the responsibility to maintain keep both places in sync onto the maintainers of this styles repo.

However, an alternative approach would be to remove that setting from the checker then and ask lesson maintainers to use an `.editorconfig`-respecting editor or to align their RStudio settings with it.

maxim-belkin:
> But puts the responsibility to maintain keep both places in sync onto the maintainers of this styles repo.

This can be solved either by adding clear comments to these lines **and/or** with a wiki.

PR #223, opened 2018-04-04
don't break words in code

fmichonneau:


assets/css/bootstrap.css
```diff
@@ -107,6 +107,7 @@ pre,
 samp {
   font-family: monospace, monospace;
   font-size: 1em;
+  hyphens: none;
```
slimlime: This is what I was referencing before, just overriding a bootstrap.css which is a _style_ file

Commit message doesn't elaborate on this one life diff.

I was just passing by when I saw the merge referenced and thought it wasn't very informative.

Possible interpretation: This merge updates the website styles so that hyphens are not visible.
It may affect the PRESENTATION of samp tag `<samp>`, if that is what is used for code samples display in the web resources  <sub>  `<code>` `<pre>` `<kbd>` </sub>

But not directly related to the 'code' as I understood it.

Edit: 
Not super important, but as Bootstrap is a standard, I thought it should be referenced and addressed in more detail before someone tries to make edits and finds *unintended behaviour*. This feels like it might be a hacky workaround just to avoid displaying hyphens that other people have accidentally left in, or to silence a code convention / enforce stylistic preference

slimlime: Not super important, but as Bootstrap is a standard, I thought it should be referenced and addressed in more detail before someone tries to make edits and finds *unintended behaviour*

maxim-belkin: @slimlime what unintended behavior are you talking about?

PR #224, opened 2018-04-09
Update repo_check.py with new GitHub labels

fmichonneau:


bin/repo_check.py
```diff
@@ -131,7 +142,7 @@ def check_labels(reporter, repo_url):
 
     overlap = set(EXPECTED.keys()).intersection(set(actual.keys()))
     for name in sorted(overlap):
-        reporter.check(EXPECTED[name] == actual[name],
+        reporter.check(EXPECTED[name].lower() == actual[name].lower(),
```
rgaiacs: Why the `lower()`?

PR #243, opened 2018-05-08
Makefile: add docker-serve target

alee:
If Docker is installed, `make docker-serve` will download and start a Jekyll container to build and serve the site with hot reloading. This binds the Jekyll webserver to localhost:4000 for security.

JEKYLL_VERSION should be synced with https://pages.github.com/versions/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 # Settings
 MAKEFILES=Makefile $(wildcard *.mk)
 JEKYLL=jekyll
+JEKYLL_VERSION=3.7.3
 PARSER=bin/markdown_ast.rb
 DST=_site
 
@@ -15,6 +16,10 @@ all : commands
 ## commands         : show all commands.
 commands :
 	@grep -h -E '^##' ${MAKEFILES} | sed -e 's/## //g'
+
+## docker-serve     : use docker to build the site
+docker-serve :
+	docker run --rm -it -v ${PWD}:/srv/jekyll -p 127.0.0.1:4000:4000 jekyll/jekyll:${JEKYLL_VERSION} make serve
 
 ## serve            : run a local server.
 serve : lesson-md


rgaiacs:
@alee Thanks for the pull request. @fmichonneau will have a look on it after CarpentryCon.

Makefile
```diff
@@ -16,6 +17,10 @@ all : commands
 commands :
 	@grep -h -E '^##' ${MAKEFILES} | sed -e 's/## //g'
 
+## docker-serve     : use docker to build the site
+docker-serve :
+	docker run --rm -it -v ${PWD}:/srv/jekyll -p 127.0.0.1:4000:4000 jekyll/jekyll:${JEKYLL_VERSION} make serve
```
maxim-belkin: I suggest using:
```
docker run --rm -it -v ${PWD}:/srv/jekyll -p 80:4000 jekyll/jekyll jekyll serve
```
and removing the `JEKYLL_VERSION` var

alee: Care to explain why?

- pinning versions to https://pages.github.com/versions/ means you'll be using the exact version of Jekyll used to render the github pages site (AFAIK)
- the port to bind to might be worth parameterizing but if you're used to Jekyll you'll be used to looking for its dev server on 4000 
- binding to localhost is in general a good security practice. If you're planning to use this to serve a network-visible Jekyll site httpd or nginx would be a better choice..

maxim-belkin: 1. With docker we have the ability to use whatever version of jekyll we wish, so why not use the latest (and the greatest) one?
2. Regarding port... using `80` would allow us to simply just type `localhost` and that's it.
> if you're used to Jekyll

Why should we care about those who are used to jekyll? (this should be read as a rhetoric question as I'm not trying to offend anyone). Why would anyone want to type extra characters (`:4000`) to see the lesson? I just want to type `localhost` and see the lesson!
3.
> binding to localhost

We do bind to localhost even if we don't specify 127.0.0.1, we just don't say it explicitly... right? I'm not an expert but in my tests I was not able to open that page anywhere else

gdevenyi: You can't bind to port 80 as a non-privileged user. Needs to be a port > 1024

maxim-belkin: 1. One can't install Docker as a non-privileged user
2. I tested `docker run --rm -it -v ${PWD}:/srv/jekyll -p 80:4000 jekyll/jekyll jekyll serve` locally and it works. Is that limitations specific to a particular version? Or am I using docker incorrectly?

alee: No offense taken, asking questions is how you learn.

1. in general it's good practice to pin your dependency versions for reproducible builds. google around and you'll find plenty of reasons but a primary one is to ensure that what you see on your local environment is what others will see in theirs (and what is being rendered on the lesson websites).
2. the real problem here is finding an unallocated port, otherwise the command will fail. Docker will magic away the privileged binding to port 80 so you can actually bind to port 80 if you want. However, some of us have real webservers already bound to port 80 serving other apps and sites. Furthermore, using the Jekyll devserver port isn't slavish abandon to a cargo cult - it's a deliberate choice to adopt the same standards documented in the [Jekyll documentation](https://jekyllrb.com/docs/usage/) and all over the interweb. 
3. This depends on your configured docker daemon settings but the default binds to 0.0.0.0 AFAIK. Of course if you're on DHCP you're only visible on local network but it's always good security practice to limit to localhost.

maxim-belkin: dependency-pinning for reproducible builds is fine. But here we would be forcing ourselves to eternally checking for the version of jekyll that GitHub uses. Reproducibility is important but so is sustainability. Unless this procedure is documented / automated, that would not be sustainable. Is there some sort of `jekyll/github` or `github/jekyll` repo?

if someone runs a webserver on a (hopefully laptop) computer used for SWC presentation... Well, first of all, I'm impressed and kudos to that person. Second of all, I think that person would know how to modify the Makefile :)

0.0.0.0 is inside of the container. on the outside it is still localhost (well, 127.0.0.1), no? At least, I don't see any `--ip` flags and I don't see how it can bind to something else.

alee: Ah I think I see where you're coming from now. I've been primarily viewing this as a tool for maintainers to check if incoming changes will break the site. In that case it seems important to maintain high fidelity with the version of Jekyll being used to render the lesson website at `swcarpentry.github.io/<foo>`. If/when the official version number changes, bump the makefile var. I appreciate the sustainability sentiment but it's misguided in my opinion - using jekyll:latest doesn't guarantee sustainability - at best it gives the false confidence of forgetting about it until something breaks. This is the same situation we'd be in with the pinned version, but at least you won't be potentially rendering things differently should Jekyll 4.0 decide to change in some backwards-incompatible way.

0.0.0.0 is on the host machine. You bind it to another IP on the host machine via the -p parameter as shown in this changeset 👀 .

These are trivial changes, so I won't belabor this discussion anymore by repeating myself - it's up to you all as a community to decide what would work best! 👍

alee: I forgot to add - ideally this is also for lesson contributors with Docker installed to more easily check if their changes are going the way they expect to minimize maintainer burden 😁 

That makes security first + sticking to Jekyll devserver conventions a bit stronger call. So IMO this is about a usable set of configurable defaults that don't break the principle of least astonishment for new users.

maxim-belkin:
@fmichonneau, please have a look at this one when you get a chance. I think it is better to merge this sooner than later to see if (other) maintainers stumble upon any issues.

fmichonneau:
Thanks all for your work on this! I'm going to merge this, and see how it works for our maintainers. Personally, I think we should provide the same full environment as the one provided by GitHub pages (with the GitHub ruby gem installed).

PR #250, opened 2018-05-15
Favicons and logos

fmichonneau:
This PR adds favicons in all sizes for broad device/browser compatibility across SWC, DC, LC and a new "CP" category. This last category is for a generic "The Carpentries" to be used for instructor training for instance.

It also fixes the path for the logos to be displayed in the navbar.


fmichonneau:
Thanks for the review Raniere. 

I don't think a loop would help much here. 
I fixed the incorrect "alt" text for the Carpentries logo and removed the unneeded `code.txt` files (sorry to have missed that!)

PR #253, opened 2018-05-17
lesson_check.py: Exit with a status of 1 if something was wrong

maxim-belkin:
To make `lesson-check` useful in PRs, it has to return non-zero status.

For now, I'm making `lesson-check-all` _permissive_, but this might change in the future.


maxim-belkin:
Thanks for merging @rgaiacs !

PR #256, opened 2018-05-18
lesson_check: get_val is a static method

maxim-belkin:
`get_val` should be a static method as it does not use `self` or `cls`

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -399,7 +399,8 @@ class CheckBase(object):
                     return False
         return True
 
-    def get_val(self, node, *chain):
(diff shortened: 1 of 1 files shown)


rgaiacs:
`get_val` is only used inside the class. Can we keep it as a non-static method to make the script simple for beginners?

maxim-belkin:
I'm not sure I understand the question. `get_val` is independent of the class it is sitting in and `@staticmethod` decorator simply highlights this fact... The other two alternatives -- instance method (has `self` as the first argument) and class method (has `cls` as the first argument) -- are not really appropriate here....

PR #278, opened 2018-05-23
Run lesson-fixme from within lesson-check

maxim-belkin:
Make `lesson-check` invoke `lesson-fixme`
`lesson-fixme` rule does not return an error, so the exit status of `lesson-check` is not impacted


maxim-belkin:
Related: swcarpentry/styles#264

fmichonneau:
Thanks @maxim-belkin. I'll let @rgaiacs merge.

PR #282, opened 2018-05-30
remove permalink that breaks asset includes

mr-c:
Will also match https://raw.githubusercontent.com/swcarpentry/lesson-example/gh-pages/CONDUCT.md


maxim-belkin:
@mr-c, please add `root: ..` under the `permalink` line instead.

rgaiacs:
This looks good to me. We should not have `root`.

PR #283, opened 2018-05-30
Match new usage of the email parameter

mr-c:
See https://github.com/swcarpentry/styles/blob/0e0f2bf4e396419a46760d0add6af054cd1c512c/_includes/lesson_footer.html#L42 and https://github.com/swcarpentry/styles/blob/0e0f2bf4e396419a46760d0add6af054cd1c512c/_includes/workshop_footer.html#L22

### DIFF
--- a/bin/boilerplate/_config.yml
+++ b/bin/boilerplate/_config.yml
@@ -8,10 +8,8 @@ carpentry: "swc"
 # Overall title for pages.
 title: "Lesson Title"
 
-# Contact.  This *must* include the protocol: if it's an email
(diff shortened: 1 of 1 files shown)


maxim-belkin:
```
 $ grep -r "site.email" *
_includes/lesson_footer.html:	<a href="mailto:{{ site.email }}">Contact</a>
_includes/workshop_footer.html:	<a href="mailto:{{ site.email }}">Contact</a>
```
so, I agree that having `site.email` equal to `mailto:lessons@software-carpentry.org` does **not** make much sense

PR #284, opened 2018-06-01
Add links.md to files in boilerplate

maxim-belkin:
Adding `links.md` to boilerplate files.

### DIFF
--- a/bin/boilerplate/_episodes/01-introduction.md
+++ b/bin/boilerplate/_episodes/01-introduction.md
@@ -9,3 +9,5 @@ objectives:
 keypoints:
 - "First key point."
 ---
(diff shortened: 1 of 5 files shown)


rgaiacs:
@BrownSarahM pointed to me that we can move the {% include links.md %} to `_layout`. Could you follow Sarah's suggestion? Thanks.

rgaiacs:
@brownsarahm I try to move `{% include links.md %}` to `_layouts` but

screencapture-localhost-4000-2018-06-03-12_05_36

maxim-belkin:
I think this PR is good as is. We can change the place where we add `links.md` later.

PR #298, opened 2018-06-15
add cp as carpentry lesson type (e.g. for instructor training)

fmichonneau:


bin/boilerplate/_config.yml
```diff
@@ -2,7 +2,7 @@
 # Values for this lesson.
 #------------------------------------------------------------
 
-# Which carpentry is this ("swc", "dc", or "lc")?
+# Which carpentry is this ("swc", "dc", "lc", or "cp")?
 carpentry: "swc"
```
maxim-belkin: would it help to add a clarifying comment as to what `cp` and, perhaps, other abbreviations mean?

fmichonneau:
thanks, that's a good point. I added the definitions. I also took advantage of this to emphasize there is no need to edit `<USERNAME>/<PROJECT>` (several lessons have changed it).

fmichonneau:
@maxim-belkin or @rgaiacs if this looks good to you, please merge.

Issue #309, opened 2018-06-22
All episodes in one page doesn't show figures


rgaiacs:
**This was reported by Cameron Macdonell**.

http://carpentries.github.io/lesson-example/aio.html doesn't show the images because the path to images is wrong.

http://carpentries.github.io/lesson-example/aio.html will have

```
<img src="../fig/file-mapping.svg" alt="Source and Destination Files">
```

that will be resolved to http://carpentries.github.io/fig/file-mapping.svg when it should be resolved to http://carpentries.github.io/lesson-example/fig/file-mapping.svg.

# Why this is happening

URL for each episode is like http://carpentries.github.io/lesson-example/XX-title/index.html.

# How to fix this

Change the URL from http://carpentries.github.io/lesson-example/aio.html to http://carpentries.github.io/lesson-example/aio/index.html.

1. Add `permalink: aio/index.html` to YAML header.
2. Add `root: ..` to YAML header.
3. Update https://github.com/carpentries/styles/blob/gh-pages/bin/lesson_check.py.

rgaiacs:
carpentries/lesson-example#220 has one alternative solution to this one. It moves `aio.md` to `_extras`.

rgaiacs:
@maxim-belkin I believe this was fixed. Could you confirm? Thanks.

fmichonneau:
this has been addressed.

PR #310, opened 2018-06-22
Replace page.root

fmichonneau:
This is a proposal to make our template more robust by not having to specify `page.root` by hand.

Instead, we use the current URL to generate automatically what used to be `page.root`.

@maxim-belkin  and @rgaiacs please review. 

If you think it's a reasonable approach, we can simplify `_config.yml` a little by removing the assignments to `root`.

This approach works well to generate the site with `make serve` or `make site`. It would solve #309 and the fact that `reference` is also broken in 9.5.2.


rgaiacs:
> This is a proposal to make our template more robust by not having to specify page.root by hand.

Looks reasonable.

You should edit https://github.com/carpentries/styles/blob/gh-pages/bin/boilerplate/_config.yml to remove references of `root`.

> It would solve #309.

It would **not** resolve carpentries/styles#310 because the "all episodes in one page" is generated by Javascript injection. You need to move that page be server at http://carpentries.github.io/lesson-example/aio/index.html.

maxim-belkin:
I merged this PR into my [personal fork](https://maxim-belkin.github.io/python-novice-inflammation) of `python-novice-inflammation` and so far I don't see images on the [all-in-one page](https://maxim-belkin.github.io/python-novice-inflammation/aio.html)

maxim-belkin:
Even though it does not solve the #309, I think this change is an improvement.
My only question is if `base` is a good name for a var.

_includes/base_path.html
```diff
@@ -0,0 +1,21 @@
...
+This is adapated from: https://ricostacruz.com/til/relative-paths-in-jekyll
+{% endcomment %}
+
+{% assign base = '' %}
+
+{% comment %}
+page.url gives the URL with a leading /
+- when the URL ends with the extension (e.g., /foo/bar.html) then we can get the depth by counting the number of / and remove - 1
+- when the URL ends with a / (e.g. /foo/bar/) then the number / gives the depth directly
+{% endcomment %}
```
maxim-belkin: I'd suggest merging this comment with the one above

_includes/all_keypoints.html
```diff
@@ -7,7 +7,7 @@ <h2>Key Points</h2>
   {% unless episode.break %}
     <tr>
       <td class="col-md-3">
-        <a href="{{ page.root }}{{ episode.url }}">{{ episode.title }}</a>
+        <a href="{{ base }}{{ episode.url }}">{{ episode.title }}</a>
```
maxim-belkin: I'd suggest calling this variable something like `lesson_url`

maxim-belkin:
I will rebase this PR

_includes/all_keypoints.html
fmichonneau: given that this variable stores repetitions of `../`, should we call it `relative_path` or `root_location`, or maybe `base_path`, I don't think that `lesson_url` is very accurate, and I agree that `base` is too vague.

maxim-belkin: oh, right, I forgot this is a relative path... I like the `root_location` and `base_path` (with a slight preference for the former). Perhaps, we should add `relative` to make it clear that this is relative path for root directory?

`relative_root_path`, `relative_root_location`, ...?

_includes/navbar.html
```diff
@@ -14,64 +14,64 @@
       {% comment %} Select what logo to display. {% endcomment %}
       {% if site.carpentry == "swc" %}
       <a href="{{ site.swc_site }}" class="pull-left">
-        <img class="navbar-logo" src="{{ page.root }}/assets/img/swc-icon-blue.svg" alt="Software Carpentry logo" />
+        <img class="navbar-logo" src="{{ base }}{% link /assets/img/swc-icon-blue.svg %}" alt="Software Carpentry logo" />
```
maxim-belkin: By the way, why does it have to be `{% link /assets/img/swc-icon-blue.svg %}` ?

maxim-belkin: Or is it an artifact from another `sed`-based work? :)

maxim-belkin: or, this is the "right way"

maxim-belkin: Link: https://jekyllrb.com/docs/templates/#link

maxim-belkin:
Shall we use
```
{{ "/assets/style.css" | relative_url }}
```
(from [here](https://jekyllrb.com/docs/templates/))

fmichonneau:
I don't think it's enough. From my understanding `|relative_url` will only append `site.baseurl` and doesn't know how deep in the folder structure the file lives.

maxim-belkin:
Thanks, @fmichonneau.
I suggest this PR is "Squashed and Merged" because of the  merge commit (d279a08)

fmichonneau:
I have tested this a little and it seems to work OK. I'll open a new issue in the lesson-example repository to remind ourselves that we should document this new variable and discourage the use of `page.root`.

brownsarahm:
I'm late to the discussion, but I have a question as to why we don't use the default jekyll variable site.baseurl ? (I used this to fix issues in the [proposal on changing exercies](https://github.com/carpentries/lesson-infrastructure/issues/21)

fmichonneau:
`site.baseurl` is used to specify the "folder" in which your site lives. If your site's URL is https://test.com/foo/, the baseurl would be `foo`. When using GitHub, `site.baseurl` is defined as `/name-of-repo`. Unfortunately this variable has no sense of how deep the page sits in the folder hierarchy of the site.

gvwilson:
Can all refs be absolute as `{{site.baseurl}}/something/somewhere/` instead of worrying about relative paths? That's worked for me in some other (simpler) projects, and as long as most of the inter-page refs are in templates, most authors don't need to worry about it.

fmichonneau:
We are trying to make the site fully functional offline so the content of `_site` can be distributed to learners where there is poor/no internet access. When using `site.baseurl`, the absolute paths don't make this possible.

With this pull requests, authors don't need to worry about where the page will live as the variable `relative_root_page` will resolve to the correct relative location, and links written as `{{ relative_root_page }}{% link path/to/file.md %}` should resolve properly.

PR #321, opened 2018-07-01
Only knit rmd files to md if they have changed

jsta:
This PR is a follow-on of [my comment](https://github.com/carpentries/styles/pull/153#issuecomment-401537578) on #153. The changes make it so that only Rmd files that have been changed are knit to md. They were motivated by the fact that the total time to build all the episodes of https://github.com/datacarpentry/r-raster-vector-geospatial is very long. This makes the existing setup difficult for local interactive development. The workflow would then be to call `make lesson-md` to build only the Rmds that have changed and `make -B lesson-md` to build all Rmds regardless of the `touch` stamp.

### DIFF
--- a/bin/knit_lessons.sh
+++ b/bin/knit_lessons.sh
@@ -4,5 +4,5 @@
 # The Makefile passes in the names of files.
 
 if [ $# -ne 0 ] ; then
-    Rscript -e "source('bin/generate_md_episodes.R')"
+    Rscript -e "source('bin/generate_md_episodes.R')" $*
 fi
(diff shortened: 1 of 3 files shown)


fmichonneau:
I have been trying this, and it looks like it's working well. Merging for further testing.

bin/knit_lessons.sh
```diff
@@ -4,5 +4,5 @@
 # The Makefile passes in the names of files.
 
 if [ $# -ne 0 ] ; then
-    Rscript -e "source('bin/generate_md_episodes.R')"
+    Rscript -e "source('bin/generate_md_episodes.R')" $*
```
maxim-belkin: I'd suggest using `"$@"` or `"$*"` instead of `$*`.

fmichonneau: thanks for the suggestion. I pushed a commit to use "$@"

Issue #161, opened 2017-08-01
Use Jekyll/Liquid include for images/figures instead of Markdown


rgaiacs:
At the moment we are using "standard" Markdown syntax to include images/figures. [Markdown syntax](https://kramdown.gettalong.org/syntax.html#images) lacks some features such as caption. I want to propose we start using Jekyll/Liquid include for the next release.

# Current behaviour

~~~
$ cat image.md 
![foo](foo.png 'foo text description')
$ kramdown image.md 
<p><img src="foo.png" alt="foo" title="foo text description" /></p>
~~~

# Propose behaviour

~~~
$ cat image.md 
{% include figure.html filename="foo.png" caption="Caption describing foo" alternative_text="foo text description" %}
$ kramdown image.md 
<figure>
  <img
  src="foo.png"
  alt="foo text description">	
  <figcaption>Caption describing foo</figcaption>
</figure>
~~~

# Required parameters

- `filename`
- `alternative_text`

# Optional parameters

- `caption`
- `license`, preferable CC-BY
- `source_url`, for any content that we didn't authored

# Backward compatibility

This will not break backward compatibility. Any lesson that didn't migrate to use include will not see any difference.

gvwilson:
Can Kramdown parse these includes to give us a list of figures? If not, how will we regenerate `_includes/all_figures.html` ?

rgaiacs:
> Can Kramdown parse these includes to give us a list of figures?

Jekyll uses Liquid for all the tags. We could change https://github.com/swcarpentry/lesson-example/blob/gh-pages/bin/markdown_ast.rb to do the tag processing replacement, see 
https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers, before kramdown do their job.

brucellino:
> how will we regenerate _includes/all_figures.html ?
I usually deal with this in a liquid loop. 
Keep a `yml` file in `_data` (_e.g._ `_data/figures.yml` ) and then loop over them in the html : 

{% for figure in site.data.figures %}
<img src="{{ figure.url }}" alt={{ figure.alt }}">
{% endfor %}

Forgive me if I'm misunderstanding something, but isn't this what you're looking for  ?

rgaiacs:
@brucellino we don't keep a list of figures on any file because if authors need to edit more than one place we have a point of failure on our workflow. For that reason, we need a script to collect all the figures "in the correct order".

brucellino:
[...]

PR #323, opened 2018-07-03
center images

fmichonneau:


maxim-belkin:
may I suggest using:
```
margin: 20px auto;
```
?
`20px` can be altered, but might be a good starting point.

fmichonneau:
yes, good idea. 20px looks good to me.

maxim-belkin:
This affects SWC logo in the top left corner (`img.navbar-logo`)

fmichonneau:
good catch, I restricted to "article"

assets/css/lesson.scss
```diff
@@ -137,8 +137,10 @@ code {
   background-color: #e7e7e7;
 }
 
-img {
-  max-width: 100%;
+article img {
+    display: block;
```
rgaiacs: I don't think that we want this one.

maxim-belkin: `display: block` is necessary for the centering to happen.

On a different matter: I think we still want to have `img { max-width: 100%; }` for all "other" images.

rgaiacs: Consider

```
$ cat f.md 
![Alt text](image.jpg)

Paragraph with ![Alt text](image.jpg) some image.
$ kramdown f.md
<p><img src="image.jpg" alt="Alt text" /></p>

<p>Paragraph with <img src="image.jpg" alt="Alt text" /> some image.</p>
```

I would say that we agree that the first image should be a block without exception. The second image doesn't necessary should be a block, for example, it is a small image of the icon that the reader should click in.

maxim-belkin: True, but I don't see the second type of usage in SWC/DC repos. Is there a need to support it at the moment?

rgaiacs: > I don't see the second type of usage in SWC/DC repos.

I don't think that we ever use it but I imagine some examples where people might want to use, e.g.

screenshot from 2018-07-05 18-10-21

> Is there a need to support it at the moment?

I want to avoid the problem of one contributor that is not familiar to CSS who might write

```
Paragraph with ![Alt text](image.jpg) some image.
```

expecting to have the image inline because (1) this is how word processors works and (2) this is the default behaviour in HTML but our lesson have the image as a block.

[Looks like](https://stackoverflow.com/questions/22179642/css-first-child-selector-including-text-nodes) that is impossible to have a CSS-only selector for paragraphs with only one image **and** not text content.

For the record, I suggested https://github.com/carpentries/styles/issues/161 so that we could have better images.

fmichonneau: My CSS skills aren't great but couldn't we add:

```css
article p > img {
  block: inherit
}
```

to cover the case of images embedded with `<p>` tags?

rgaiacs:
@fmichonneau after think about this, my recommendation is to require authors to use a class. For example, to use

```
![Alt text](image.jpg)
{: .figure}
```

instead of

````
![Alt text](image.jpg)
```

kramdown ouput will be

```
$ cat f.md 
![Alt text](image.jpg)
{: .figure}

Paragraph with ![Alt text](image.jpg) some image.
$ kramdown f.md 
<p class="figure"><img src="image.jpg" alt="Alt text" /></p>

<p>Paragraph with <img src="image.jpg" alt="Alt text" /> some image.</p>
```

And the addition to the CSS would be

```
article img.figure {
    display: block;
}
```

This will ensure that figures that we don't want to be centered aren't.

fmichonneau:
Given that we currently don't have inline images I think it makes more sense to create a class for them if we end up needing it. Let's merge and test this thoroughly before the next release.

Issue #168, opened 2017-08-31
Change font because 1 and l are similar.


rgaiacs:
@dakane1 opened https://github.com/swcarpentry/shell-novice/issues/612 saying that

> I had a number of issues distinguishing 1 (ones) from the letter l. When I mentioned it to a few people they said they were having the same problem.

# Screenshot

![screenshot from 2017-08-31 19-29-21](https://user-images.githubusercontent.com/1506457/29939195-fd4c6740-8e82-11e7-9472-5a130e2a8cc5.png)

# Technical details

If you are using a high resolution display (HIDPI) like the retina display in some MacBooks, the QHD+ in some laptops or (you are the luck one with) the 4k monitor you may have some issues due many factors that some fonts doesn't scale properly and the text looks **super** small making hard to read and distinguish some symbols that are similar. **If this is your case, sorry but we can't do much for you.**

The CSS related information about the font that we are using in the main body of the lesson is

~~~
font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
font-size: 14px;
line-height: 1.42857143;
color: #333;
~~~

Because of my completely lack of design skills, I'm open to suggestions about what we should use as `font-family`. **As long as you suggest a open source font**, I will not oppose to make the change.

dakane1:
I had actually printed the lessons and was not reading them on a screen, not sure if that information helps or not.  http://swcarpentry.github.io/shell-novice/aio/.  I can ask a web designer I know to see if they have any font recommendations.

rgaiacs:
@dakane1 Thanks for the information. It is very useful.

For the main text, the only difference of the CSS used for print is

~~~
color: #000 !important;
~~~

which enforce a darker colour.

> I can ask a web designer I know to see if they have any font recommendations.

Thanks in advance.

rgaiacs:
@dakane1 did you have the opportunity to ask for suggestion from your web designer friend?

dakane1:
Sorry, not yet.  Start of the quarter has been crazy busy but I will ask after OA Week, which is next week.

fmichonneau:
Here are some options:

# Cabin

License: Open Font License

`1Il0O`: 
![screenshot from 2018-06-15 09-50-34](https://user-images.githubusercontent.com/5502922/41471417-950b89ea-7081-11e8-8014-9731a753ece1.png)

Preview:
[...]

PR #324, opened 2018-07-03
update fonts

fmichonneau:
This addresses #168.

In addition to using Ubuntu for the font of the text, I kept the default font for the navigation and the footer, tweaked slightly the size of the fonts, and use Lato for the headers. 

Before:
screenshot from 2018-07-03 16-38-30

After:
screenshot from 2018-07-03 16-38-44


maxim-belkin:
This implementation lacks bold/italics/bold italics variants of Ubuntu. Also, TBH, I'm not sure I like `Lato` font more than Ubuntu for titles.

[Here](https://gist.github.com/maxim-belkin/ced2b8c8c6e8c148d147f9352d01f9dd
) is a CSS I created for my other project. For that, I had to download Ubuntu fonts from [here](https://assets.ubuntu.com/v1/fad7939b-ubuntu-font-family-0.83.zip), unpack them, and create that CSS. It defined Ubuntu and Ubuntu Mono.

rgaiacs:
> In addition to using Ubuntu for the font of the text, I kept the default font for the navigation and the footer, tweaked slightly the size of the fonts, and use Lato for the headers.

Why keep the font for navigation and use another for headers? My reading of your sentence is "navigation and headers are second class citizens in our lesson that don't need to be accessible to everyone."

rgaiacs:
@fmichonneau any update on this one?

fmichonneau:
[The Carpentries remote theme](https://github.com/carpentries/carpentries-theme) uses Ubuntu (and has done so since May 2019). I haven't heard compliments or complaints about it. Maybe it's time to bring this proposed change to the maintainers?

PR #325, opened 2018-07-03
Fix footer and copyright notice

fmichonneau:
Make footer a little tidier, and replace the word "Copyright" with "CC-BY"

Before:
screenshot from 2018-07-03 17-11-01

After:
screenshot from 2018-07-03 17-11-15

### DIFF
--- a/_includes/lesson_footer.html
+++ b/_includes/lesson_footer.html
@@ -40,12 +40,9 @@
 	<a href="{{ site.github.repository_url }}/blob/gh-pages/CITATION">Cite</a>
 	/
 	<a href="mailto:{{ site.email }}">Contact</a>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col-md-12" align="center">
-      Using <a href="https://github.com/carpentries/styles/">The Carpentries style</a>
-      version <a href="https://github.com/carpentries/styles/releases/tag/v9.5.2">9.5.2</a>.
+	<br>
+	Using <a href="https://github.com/carpentries/styles/">The Carpentries style</a>
+	version <a href="https://github.com/carpentries/styles/releases/tag/v9.5.2">9.5.2</a>.
     </div>
   </div>
 </footer>
--- a/_includes/lesson_footer.html
+++ b/_includes/lesson_footer.html
@@ -5,22 +5,22 @@
   <div class="row">
     <div class="col-md-6 copyright" align="left">
 	{% if site.carpentry == "swc" %}
-	Copyright &copy; 2018–{{ 'now' | date: "%Y" }}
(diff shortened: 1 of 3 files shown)


assets/css/lesson.scss
```diff
@@ -156,7 +156,6 @@ img {
 footer .copyright,
 footer .help-links
 {
-    font-size: 18px;
```
maxim-belkin: this could be: `font-size: inherit;`

maxim-belkin:
This replaces Copyright with License ([link](http://www.majordojo.com/2010/07/license-vs-copyright.php)). Is that what "The Carpentries" want?

maxim-belkin:
Looks good, btw.

fmichonneau:
I think we want to convey more clearly that our lessons are licensed under CC-BY than it currently is. The copyright is still here by having the (c) before The Carpentries. 

@tracykteal could you please confirm this looks good to you?

tracykteal:
Good idea to more clearly linking to the CC-BY. This looks good with both the license link and the copyright.

_includes/lesson_footer.html
```diff
@@ -5,22 +5,22 @@
   <div class="row">
     <div class="col-md-6 copyright" align="left">
 	{% if site.carpentry == "swc" %}
-	Copyright &copy; 2018–{{ 'now' | date: "%Y" }}
+	<a href="{{ site.cc_by }}">CC-BY</a>  &copy; 2018–{{ 'now' | date: "%Y" }}
```
rgaiacs: I think the use of `&copy'` here is inappropriate, see https://creativecommons.org/choose/.

_includes/lesson_footer.html
```diff
@@ -40,12 +40,9 @@
 	<a href="{{ site.github.repository_url }}/blob/gh-pages/CITATION">Cite</a>
 	/
 	<a href="mailto:{{ site.email }}">Contact</a>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col-md-12" align="center">
-      Using <a href="https://github.com/carpentries/styles/">The Carpentries style</a>
-      version <a href="https://github.com/carpentries/styles/releases/tag/v9.5.2">9.5.2</a>.
+	<br>
```
rgaiacs: I prefer to have this as `col-md-12` instead of under the other links as this is not part of the same "semantic" element.

_includes/lesson_footer.html
maxim-belkin: I think Raniere is right: it should be something like
```
The Carpentries &copy; 20XX - 20XX
CC-BY...
```

_includes/lesson_footer.html
maxim-belkin: Although this is a different semantic element, it is not so important to be in the middle of a page. Perhaps, it could be combined with the license block.

PR #326, opened 2018-07-03
space out callout elements better

fmichonneau:
I find the different parts of the content of the callout boxes crowded. Attempting to give them a little more room.

Before:

screenshot from 2018-07-03 17-48-14

After:
screenshot from 2018-07-03 17-46-36

PR #345, opened 2018-08-22
Code/callout/prereq/etc blocks: set left and right margins to 0

maxim-belkin:
Setting `margin: 15px` causes all blocks (Prereq, code, etc) to be misaligned with the main text. 
This PR sets the left and right margins to 0.

Related commit: 80d6638a

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -31,7 +31,7 @@ $color-testimonial: #fc8dc1 !default;
 @mixin cdSetup($color) {
     color: $color;
     border-left: solid 5px $color;
-    margin: 15px;
+    margin: 15px 0;
     border-radius: 4px 0 0 4px;
 }
 
@@ -84,7 +84,7 @@ $codeblock-padding: 5px !default;
   border-radius: 4px;
   padding-bottom: $codeblock-padding;
 
-  margin: 15px;
+  margin: 15px 0;
 
   h2 {
     padding-top: $codeblock-padding;


fmichonneau:
This was actually intended when I created #326. I think it helps structure and separate the boxes from the text better. Setting margins to 0 also crowds the edges of the boxes on the right sides.

Compare:

screenshot from 2018-08-24 14-26-49

and

screenshot from 2018-08-24 14-27-17

fmichonneau:
I propose we revert this PR.

maxim-belkin:
1. Right margin is not necessary at all, IMO.
2. Left margin might be fine. However, the value of 15px makes it look as if it is unintentionally misaligned. See code blocks and callout boxes on [this page](https://swcarpentry.github.io/python-novice-inflammation/01-numpy/index.html).

I'd be fine with separating them a bit more along the vertical axies rather than adding left/right margins.

fmichonneau:
If 15px makes it look unintentionally misaligned, maybe we need a larger value because that was definitely my intention :smile: 

I think the way it renders looks much better than having the boxes takes the full width of the pages.

maxim-belkin:
Yes I thought so it at first and tried several values - didn't like any of them (the last margin I tried was 50x `margin: 15px 0px 15px 50px;`). I think margins work fine if there is a significant amount of text around these boxes. When the amount of text is not sufficient - it seems a tad weird. That' about code blocks. As for the callout boxes -  I think horizontal margins are not necessary at all.

maxim-belkin:
I think this is an example of a style choice that should be discussed among the group of maintainers.

PR #349, opened 2018-08-24
use github metadata to identify branch to use when editing files

fmichonneau:
instead of hard-coding the name of the branch to use when editing files on GitHub, here we use the GitHub metadata to identify the default branch for the repository. This is needed as we now have several repositories that don't use gh-pages as the default branch: it is only used to host and serve the files as they are generated by scripts running on Travis. Files live on master. This is used in production at https://datacarpentry.org/r-raster-vector-geospatial

For citation and contributing, we link to the files from the branch hosting the files.


maxim-belkin:
Wouldn't it be easier to add an extra parameter to `_config.yml`, something like `default_branch` ?

maxim-belkin:
On the other hand, this solution is automated, so has its benefits

maxim-belkin:
How does this solution work offline?

rgaiacs:
> This is needed as we now have several repositories that don't use gh-pages as the default branch

Why do we don't enforce `gh-page`?

> How does this solution work offline?

No. This feature never worked oflline.

_includes/lesson_footer.html
```diff
@@ -30,17 +30,17 @@
     <div class="col-md-6 help-links" align="right">
 	{% if page.source %}
 	{% if page.source == "Rmd" %}
-	<a href="{{site.github.repository_url}}/edit/{{ default_branch }}/{{page.path|replace: "_episodes", "_episodes_rmd" | replace: ".md", ".Rmd"}}">Edit on GitHub</a>
+	<a href="{{repo_url}}/edit/{{ default_branch }}/{{page.path|replace: "_episodes", "_episodes_rmd" | replace: ".md", ".Rmd"}}">Edit on GitHub</a>
 	{% endif %}
 	{% else %}
-	<a href="{{site.github.repository_url}}/edit/{{ default_branch }}/{{page.path}}">Edit on GitHub</a>
+	<a href="{{repo_url}}/edit/{{ default_branch }}/{{page.path}}">Edit on GitHub</a>
```
maxim-belkin: It might make sense to display this only when the page is rendered "online"

_includes/lesson_footer.html
```diff
@@ -30,17 +30,17 @@
...
 	{% endif %}
 	/
-	<a href="{{ site.github.repository_url }}/blob/{{ site.github.source.branch }}/CONTRIBUTING.md">Contributing</a>
+	<a href="{{ repo_url }}/blob/{{ source_branch }}/CONTRIBUTING.md">Contributing</a>
```
maxim-belkin: Why not simply `/CONTRIBUTING.md` ? This is a relative path, relative to the root of the website being rendered, so it should work regardless of the branch, repo_url, etc

fmichonneau: We have been pointing to the CONTRIBUTING file on GitHub instead of a Jekyll-rendered version. I think it makes sense to send possible contributors to GitHub directly, and it actually may make more sense to also point to the default branch here as well.

maxim-belkin: > I think it makes sense to send possible contributors to GitHub directly

Why jump back and forth? I think all but the `Edit on GitHub` link should keep viewers on the rendered website. Information will be the same but presented in a better (rendered) way.

_includes/lesson_footer.html
maxim-belkin: This might create a loophole: if something does not work in the rendered version, it might not indicate a problem.

I guess, it would make more sense if this `Edit on GitHub` link points to GitHub no matter where it is rendered.

fmichonneau: This is out of the scope of the problem that this PR is addressing. I opened a separate issue (#352) to keep track of it.

_includes/lesson_footer.html
fmichonneau: I opened a separate issue to address this (#351)

fmichonneau:
@rgaiacs in the geospatial curriculum, the default branch is master because gh-pages holds the output generated by the compilation of the Rmd files.

fmichonneau:
any objections to merge this?

_includes/gh_variables.html
```diff
@@ -0,0 +1,45 @@
+{% comment %}
+When rendering websites locally, `site.github.url` doesn't get resolved
+properly unless GitHub PAT are set up and available in the environment.
```
maxim-belkin: `GITHUB_PAT` -- GitHub Personal Access Token

_includes/gh_variables.html
```diff
@@ -0,0 +1,45 @@
...
+This leads to warnings and errors when trying to serve the site locally.
+To work around this, we use the `jekyll.environment` variable which is set to
+`development` when rendering the site locally, and set to `production` on
+GitHub where `site.github.url` will be defined.
```
maxim-belkin: is defined

_includes/gh_variables.html
```diff
@@ -0,0 +1,45 @@
...
+{% endcomment %}
+{% assign default_branch = repo_info[0].default_branch %}
+
+{% comment %}
+Other variables requested by the template
+{% endcomment %}
+{% assign repo_url = site.github.repository_url %}
+{% assign search_domain_url = site.github.url %}
+{% assign project_title = site.github.project_title %}
+{% assign source_branch = site.github.source.branch %}
```
maxim-belkin: what is the difference between `source_branch` and `default_branch`?

fmichonneau:
default_branch is the default branch for the repository (usually master but gh-pages in many of our repositories), source_branch is the branch used to serve the GitHub pages website (in our case almost always gh-pages).

maxim-belkin:
Sounds good, but do we need to define `repo_url`, `search_domain_url`, and `project_title` in this PR?

Also, I'd suggest to use spaces around text in `{{ text }}`

fmichonneau:
I updated the comments with your suggested edits Maxim. While I agree with your stylistic suggestion of having spaces between the curly brackets and the variable name, I suggest we take care of this for the repository in a separate PR.

As for the variables, I assume you mean for the case of local rendering. I'm not quite sure what would be appropriate values and we can try to look for solution in #352.

maxim-belkin:
> As for the variables, I assume you mean for the case of local rendering.

Apologies for not being clear. What I meant was that we don't need these other variables -- they're just a convenience factor... unlike `default_branch` which is actually extracted from a blob of information.

fmichonneau:
We actually do need them because if we use for instance `site.github.url` directly, then local rendering using `jekyll serve` fails as the GitHub metadata is not available.

maxim-belkin:
Ah! Alright then! Thanks for the explanation!

PR #350, opened 2018-08-27
Reflect Apple's rebranding

katrinleinweber:
See diff & [apple.com/macos](https://www.apple.com/macos/) ;-)


fmichonneau:
Thanks @katrinleinweber!

Issue #356, opened 2018-09-24
Lesson check error


maxim-belkin:
`make lesson-check` now returns an error:

```
$ make lesson-check
_includes/links.md:1 not valid reference:
{% include gh_variables.html %}
make: *** [lesson-check] Error 1
The command "make lesson-check" exited with 2.
```

https://travis-ci.org/swcarpentry/python-novice-inflammation/builds/432404616?utm_source=github_status&utm_medium=notification

maxim-belkin:
My understanding is that `gh_variables.html` was included into `links.md` to aid resolving  `repo_url` and `source_branch`. However, resolution happens just fine even without that inclusion.

PR #358, opened 2018-10-02
Remove gh-variables from links.md

maxim-belkin:
See description in carpentries/styles#356

This blocks execution of `make lesson-check`, so I'm marking this as high priority.


maxim-belkin:
Thanks, François!

PR #360, opened 2018-10-15
Adding Gemfile and Gemfile.lock

maxim-belkin:
Use Gemfile and Gemfile.lock files to make sure all carpentries' maintainers are using the same gems (such as `jekyll`)


maxim-belkin:
I'll improve this PR and will reopen it later.

PR #370, opened 2018-12-02
Add Carpentries, LC site, issues

libcce:


maxim-belkin:
We could include and use `links.md` here...

fmichonneau:
I opened #377 to remember to add links.md to this file.

PR #371, opened 2018-12-02
Add LC to instruction material section

libcce:


maxim-belkin:
I wonder if we should be using "The Carpentries" to refer to SWC + DC.

rgaiacs:
> I wonder if we should be using "The Carpentries" to refer to SWC + DC.

"The Carpentries" is Software Carpentry + Data Carpentry + Library Carpentry.

PR #381, opened 2018-12-07
links.md: sync with downstream (lesson-example)

maxim-belkin:
CC @rgaiacs and @fmichonneau since these links were added by both of you in carpentries/lesson-example


fmichonneau:
thanks!

PR #382, opened 2018-12-12
Rework the Makefile. Add Gemfiles and required.mk

maxim-belkin:
1. [New] Gemfile: use `gem 'github-pages'`, not just `jekyll` and `kramdown`. Specifies the minimum version of Ruby (equals to that on GitHub).
~1. [New] Gemfile.lock that specifies all gem used by GitHub engine exactly.~
1. [New] required.mk that contains all definitions used in the Makefile
1. [Reworked] Makefile
    * check for tool existence before trying to run (almost) everything
    * new 'plumbing' commands to help install required tools


maxim-belkin:
I'd like to ask @gcapes to have a look at it.

rgaiacs:
@maxim-belkin This looks fantastic.

fmichonneau:
Thanks a lot for this!

Why do you think we should include the `Gemfile.lock` in the repository or hardcode the version of Jekyll that GitHub uses?

maxim-belkin:
Thanks for the feedback!

Regarding Gemfile.lock: I don't feel strongly about including it and I see that you didn't include it in https://github.com/datacarpentry/r-intro-geospatial/commit/e825071b4ca86b7fd96abd4e1ee505fdcf06c2c9 and added it to `.gitignore` a few months later in https://github.com/datacarpentry/r-intro-geospatial/commit/a0247512e168c8a29fd6c6926fca086d03de1b21 but it looks like having this file can have some benefits.

**If we don't include Gemfile.lock:**  Because this file is generated upon `bundle install`, we have to add it to `.gitignore` as you did in r-intro-geospatial in order to (a) keep confusion to a minimum and (b) keep it from being accidentally added later down the road. In this scenario, once generated by the user, this file is most likely to remain unchanged.

**If we include it:** First of all, why? 
1. I found two places that recommend checking it into version control: `bundle install --help` and [here](https://bundler.io/v1.7/rationale.html#checking-your-code-into-version-control). "Use cases" described there are not super appealing. Therefore, see the next point.  
2. People that don't have Ruby installed will want to use Docker... And it is possible to extract the required version of jekyll from Gemfile.lock. I haven't implemented it (yet) (but I will) and this is why you see the hard-coded value in `required.mk`. If we don't include `Gemfile.lock` - we have to have a hard-coded value for the jekyll version.

required.mk
```diff
@@ -22,8 +22,12 @@ PYTHON ?= $(shell which python3)
 RSCRIPT ?= $(shell which Rscript)
 
 # Jekyll version on GitHub
-# Sync with https://pages.github.com/versions/
-GITHUB_JEKYLL_VERSION = 3.7.4
+ifneq (, $(wildcard ./Gemfile.lock))
+  GITHUB_JEKYLL_VERSION := $(shell sed -n '/jekyll\ (=.*)/s|.*(= \(.*\))|\1|p' Gemfile.lock)
```
maxim-belkin: Here, I'm using the line in `Gemfile.lock` that reads
```
      jekyll (= 3.7.4)
```
There are other lines where jekyll is listed. I use the one above because it is a direct descendant of `github-pages` gem (Line 36 in Gemfile.lock)

rgaiacs:
@maxim-belkin Looks good to me.

maxim-belkin:
I'm going to split this PR into multiple smaller ones that should be easier to process.

PR #384, opened 2018-12-12
lesson_check.py: require _config.yml too

maxim-belkin:
I think we should include `_config.yml` to the list of required files...


alexwhan:
This list is only for *markdown* files. This change has therefore broken `lesson-check.py`. There is already a function specifically to to check `_config.yml` (`check_config`).

maxim-belkin:
This list is not for markdown files only but this commit indeed broke `lesson_check` because the code that works with this list provides information about markdown files only.

`check_config` does not check whether `_config.yml` file exists, but handles its absence correctly (errors out).

PR #386, opened 2019-01-23
Clarify wording of PR & issue templates

gcapes:
I see lots of contributions where none of this placeholder text has
been deleted, but also quite a lot where the first line has been deleted,
but the 'ext below' hasn't (exactly in accordance with the current
instructions).


gcapes:
It won't do anything for the first problem of people not seeing/reading the text, but it might help with the second group who have followed the instructions to the letter :)

PR #390, opened 2019-03-01
Colorize all boxes

maxim-belkin:
Use the same colors but apply 12.5% opacity.
Solution boxes use 30% opacity

image

image

image

image

image

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -129,6 +129,17 @@ $codeblock-padding: 5px !default;
 font-size: 18px;
 }
 
+.challenge    { background-color: #eec27520; }
+.callout      { background-color: #f4fd9c20; }
+.challenge    { background-color: #eec27520; }
+.checklist    { background-color: #dfd2a020; }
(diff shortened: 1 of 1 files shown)


libcce:
Have you tested it with http://wave.webaim.org/?

maxim-belkin:
Nope. I've added this CSS to my fork of `python-novice-inflammation`, so will test it now

maxim-belkin:
This tool seems to be happy. At least, no complaints related to these changes

raynamharris:
Ah, okay. It took me a minute to realize that the big change was from a white background _inside_ the box to a colored background. I think this works. 

Another accessibility testing plugin that I like is http://khan.github.io/tota11y/

maxim-belkin:
Interesting tool! thank you for the pointer! Tested -- no issues related to proposed changes.

fmichonneau:
LGTM! Thanks

maxim-belkin:
Thank you!

PR #391, opened 2019-03-06
Improve contrast of code elements

maxim-belkin:
Tool called `tota11y` that @raynamharris mentioned in carpentries/styles#390 reports that contrast of our code elements is very low. This PR fixes this using the suggestion from that tool. Here, I'm not changing the gray background but rather darken the blue color used to display text.

### Before
image

### After

image

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -151,7 +151,7 @@ blockquote p {
 code {
   white-space: nowrap;
   padding: 2px 5px;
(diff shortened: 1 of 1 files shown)


raynamharris:
Excellent! I like the improved contrast.

ACharbonneau:
Neat :)

fmichonneau:
thanks for this!

PR #393, opened 2019-03-11
Prefix code blocks with language names

maxim-belkin:
## Regular code block (undefined and R)
image

image

## Code block + Output
image

## Code block + Error message
image

## Code block in a callout box
image

## Code and output block in a solution box
image


fmichonneau:
thanks @maxim-belkin!

PR #394, opened 2019-03-12
Fix contrast of links in challenge blocks

maxim-belkin:
Tota11y reports that links in our challenge blocks have a
contrast ratio of 4.29, which is lower than their minimum
of 4.5. This PR implements the suggestion that their tool
provides by changing the color or links from #337ab7 to #2876b5.


maxim-belkin:
## Before
image

## After
image

fmichonneau:
:+1:

PR #396, opened 2019-04-18
Avoid inconsistent grammar

katrinleinweber:
See diff ;-)

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -342,7 +342,7 @@ class CheckBase:
                 n > MAX_LINE_LEN) and (not l.startswith('!'))]
             self.reporter.check(not over,
                                 self.filename,
-                                'Line(s) are too long: {0}',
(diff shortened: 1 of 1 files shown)


bin/lesson_check.py
```diff
@@ -342,7 +342,7 @@ def check_line_lengths(self):
                 n > MAX_LINE_LEN) and (not l.startswith('!'))]
             self.reporter.check(not over,
                                 self.filename,
-                                'Line(s) are too long: {0}',
+                                'Line(s) too long: {0}',
```
katrinleinweber: Alternatively, we could expand this to `Line is / Lines are ...` but IMHO this is sufficient to avoid the inconsistency `Line(s) are ...: 1`

maxim-belkin:
Thank you, Katrin!

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #397, opened 2019-04-18
util.py: hot fix for YAML loader

maxim-belkin:
Urgent fix, see https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

@fmichonneau @rgaiacs


fmichonneau:
thanks Maxim. Should we evaluate whether we need to push this across all repositories?

maxim-belkin:
I think we should: someone could exploit our Travis integration to do some calculations ("mining") on Travis machines...

PR #398, opened 2019-05-07
make title show episode title first

maneesha:
By showing the episode title first, we can easily see what episode names when we have multiple tabs open.  For example:

**Before:**
image

**After:**
image

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -25,7 +25,12 @@
 	<script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
 	<script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
 	<![endif]-->
-    <title>{{ site.title }}{% if page.title %}: {{ page.title }}{% endif %}</title>
+
+  <title>
+  {% if page.title %}{{ page.title }}: {% endif %}
+  {% if site.title %}{{ site.title }}{% endif %}
+  </title>  
+
   </head>
   <body>
 


maneesha:
Note this does not automatically propagate to all the lesson repos. Need to work with maintainers on this.

maxim-belkin:
Hi Maneesha,

I like the idea of showing page title first. However, I'm not sure about the symbol that separates site and page titles. Also, `site.title` (lesson title) can be shown unconditionally because it is supposed to be set in `_config.yml`

_layouts/base.html
```diff
@@ -25,7 +25,12 @@
 	<script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
 	<script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
 	<![endif]-->
-    <title>{{ site.title }}{% if page.title %}: {{ page.title }}{% endif %}</title>
+
+  <title>
+  {% if page.title %}{{ page.title }}: {% endif %}
+  {% if site.title %}{{ site.title }}{% endif %}
```
maxim-belkin: I suggest merging the two lines and displaying `page.title` unconditionally.
```
{% if page.title %}{{ page.title }}: {% endif %}{{ site.title }}
```

maxim-belkin:
I've just noticed that you created a branch in this repository. I would recommend using personal forks for Pull Requests:
https://reflectoring.io/github-fork-and-pull/
https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

maneesha:
@fmichonneau is reviewing; likely OK to include both conditionals to account for off cases when `site.title` may not be set.

_layouts/base.html
maneesha: Assuming you mean "displaying `site.title` unconditionally."

maxim-belkin: yes, site title, sorry for the confusion. Site title must be set (in the `_config.yml`)

maneesha:
@maxim-belkin  I don't want to continue this conversation here as it's not relevant specifically to this project, but note that forks are generally used for divergent projects, while branches are used by collaborators to contribute to the same project.

NelleV:
@maneesha That's not an accurate representation of how people use forks nowadays. That used to be true for SVN, but not for git now that github exists. You definitely should be using a fork to contribute back to project hosted on github, and avoid pushing branches to the main repository even if you have write access to them.

PR #406, opened 2019-05-29
Improve all in one

fmichonneau:
* it seems that to make all in one more robust, we need its URL to have the same depth as the episodes. Using the permalink `/aio/index.html` allows us to do this.
* Having the code in a file in `_includes` makes it easier to update its content when maintainers want to update the version of the template. 
* I'm also wondering if we could use a more explicit name:
  - `all`
  - `combined`
  - `all_episodes`


maxim-belkin:
>     * Having the code in a file in `_includes` makes it easier to update its content when maintainers want to update the version of the template.

This is, actually, a nice idea that we can use the same approach for "figures.md" (and any other content that is shared across all repos).

maxim-belkin:
1. Can we move `bin/boilerplate/aio.md` to the root folder?
2. Shall we add a note (to both `aio.md` files) that there is no need to modify them.
3. Shall we call `aio.md` in `_includes` directory something like `aio-base.md` (or `-core` or `-something`)?

fmichonneau:
> 1. Can we move `bin/boilerplate/aio.md` to the root folder?

Yes we can. I think that `aio.md` shouldn't be in `boilerplate` though. `boilerplate` should only include files that are intended to be customized by the lesson maintainers. But it's something I'd like to revisit soon as we are planning the transition to using remote themes in more detail.

> 2. Shall we add a note (to both `aio.md` files) that there is no need to modify them.

Sure.

> 3. Shall we call `aio.md` in `_includes` directory something like `aio-base.md` (or `-core` or `-something`)?

that sounds good. How about `aio-body.md`? or `aio-content.md`, or maybe `aio-script.html`?

maxim-belkin:
> I think that `aio.md` shouldn't be in `boilerplate`

So, why don't we move it from boilerplate to the root in this PR? :)

maxim-belkin:
I like `aio-script`, `aio-core`, and `aio-base`.

fmichonneau:
@maxim-belkin could you please merge if it looks good to you?

PR #409, opened 2019-06-01
lesson-check: exclude aio.md, fix read_references

maxim-belkin:
- Don't check metadata in `aio.md` (it no longer has to be empty due to
recent changes).
- Make `read_references` skip `{% include ... %}` lines


fmichonneau:
thanks Maxim!

PR #412, opened 2019-06-17
bump docker-serve jekyll version

alee:
https://pages.github.com/versions/ now reports 3.8.5 as the official jekyll version


fmichonneau:
thanks!

PR #414, opened 2019-06-18
Callout (and other) blocks: proper font size, margins

maxim-belkin:
1. Font size in callout (and other) blocks was recently changed to 17.5px which appears to look strange. In this pull request we inherit font size (from the parent)
2. Margins in keypoint boxes are -15px, resulting in slightly different offset from the left side when compared to the code blocks. By setting margin to -10px, we make offsets in keypoint boxes and code blocks identical.


fmichonneau:
thanks, it looks good to me.

maxim-belkin:
Thanks, François!

PR #416, opened 2019-06-20
Use 1px borders: fix Google Chrome & Edge

maxim-belkin:
See https://github.com/swcarpentry/git-novice/pull/662#issuecomment-499555724

CC @kekoziar @fmichonneau


fmichonneau:
thanks!

PR #422, opened 2019-07-31
Non-numeric episode ordering

### LINKED ISSUE #236: Using explicit TOC instead of numbered filenames
Our use of numbered filenames (e.g. `_episodes/01-blah.md`) has always annoyed me, since changing the order of episodes or inserting/removing episodes required renaming multiple files that haven't actually been altered. I've therefore been experimenting with a template that puts the permalinks of the episodes in `_course.yml` - the result so far is in <https://github.com/gvwilson/js-vs-rc>, and its `CONTRIBUTING.md` file has design notes. I'm finding it much easier to use during course development; is it something that the Carpentries would be interested in looking at for their next rev?

tobyhodges:
Fixes #236 

This PR adjusts the lesson template to allow ordering of episodes & extras to be determined from a field, `contents`, in `_config.yml` instead of by numeric prefix in the filename. It draws heavily from the setup of the old Jekyll template for [js4ds](https://github.com/software-tools-in-javascript/js4ds/) by @gvwilson, as he suggested in the issue referenced above.

#### Summary of Changes

- `_includes/navbar_section.html`: a layout section that builds a dropdown list from the navigation bar based on the `contents` field of `_config.yml`. The only new file introduced in this PR. **Includes troubling workarounds for adding the links to "Reference" or "All-In-One" based on whether the list being built is for Extras or Episodes** - if anyone has any better ideas for how to address this, I'd love to hear them!
- `_includes/navbar.html`: adjusted to make use of `navbar_section.html`.
- `_includes/syllabus.html`: adjusted to list episodes based on the order defined in `contents` field of `_config.yml`.
- `_includes/episode_navbar.html`: added logic to link to the correct episode with the "previous" & "next" buttons. **This seems to work but is pretty ugly** - again I would appreciate another pair of eyes reviewing this and finding ways to make it more understandable and/or efficient.
- `bin/boilerplate/_config.yml`: added the `contents` field.
- `bin/boilerplate/01-introduction.md`: renamed to `bin/boilerplate/introduction.md`
- `bin/lesson_initialize.py`: remove numeric prefix from the created episode file.

#### Example Lesson

See an example lesson built with these changes here: https://tobyhodges.github.io/non-numeric-ordering-test/ 

(The `gh-pages` branch for this lesson includes a few other commits to add more episodes and check that the build is working as intended - check the commit history of [the corresponding repository](https://github.com/tobyhodges/non-numeric-ordering-test).)


tobyhodges:
Oops! Realised I forgot about the navigation buttons for next/previous episode. Will try to ddress that now.

tobyhodges:
Ok, added functionality for "previous" and "next" buttons in the episode navigation bar (bottom of episode pages). I've updated my original comment above to include mention of the changes to `_includes/episode_navbar.html`.

_includes/episode_navbar.html
```diff
@@ -12,8 +12,27 @@
...
+          {% assign prev = epi %}
+        {% endif %}
+      {% endfor %}
+      {% for episode in site.episodes %}
+        {% if prev_epi == episode.slug %}
+          {% assign prev_url = episode.url %}
+          {% break %}
+        {% endif %}
+      {% endfor %}
+      {% if prev_url %}
```
maxim-belkin: I think it would be better to move this to a separate file and then be included in this file... Just like `base_path.html` above.
@fmichonneau, please 👍  if you agree.

maxim-belkin:
I'm 👍  on the idea **IF** lessons can continue to use numbered episodes.
I will do a more in-depth review if that's the case.

tobyhodges:
This is a good point. It’s important not to force breaking changes on the
existing lesson maintainers. I’ll update the PR accordingly late today or
tomorrow.

tobyhodges:
This now works for both numbered episodes and for non-numbered episodes where the order is defined in the `contents` field of `_config.yml`. I moved the logic for fetching the correct URLs for previous and next episodes to a new file, as suggested by @maxim-belkin above.

- Example lesson built with non-numbered episodes: https://tobyhodges.github.io/non-numeric-ordering-test/index.html
- Example lesson built with numbered episodes: https://tobyhodges.github.io/numbered-test-build/index.html

I'm still unhappy with some of the Liquid control flow, which I'm sure could be improved by someone more familiar with it than I. Also for discussion: what do you want to be the *default* way of doing things in future? As it stands, I've changed the boilerplate and behavior of `bin/lesson_initialize.py` to create non-numbered episodes.

_includes/nav_urls.html
```diff
@@ -0,0 +1,51 @@
+{% comment %}
+Get the appropriate URLs for the previous and next episodes,
+if they exist, based on whether the episode order is defined
+in site.contents or determined by numbered filename prefix.
+
+End result is two variables: prev_url and next_url.
+{% endcomment %}
```
maxim-belkin: Could you please add description of `page.slug` here?  Perhaps, with a link to https://jekyllrb.com/docs/permalinks/ ?

tobyhodges: Thanks for the advice. I've added comment notes on this to `_includes/nav_urls.html` and `bin/boilerplate/_config.yml`

tobyhodges:
Another point to consider in relation to this PR: if merged, the changes would need further documentation in http://carpentries.github.io/lesson-example/03-organization/index.html. I'm happy to put in PRs for that also, but won't start working on anything until this PR is resolved.

mr-c:
@maxim-belkin Thank you for your review. Any additional comments? We'd like to use this for the CWL user guide (which is based upon this repo).

maxim-belkin:
Thanks for the ping, @mr-c and apologies for dropping the ball, @tobyhodges.

I began reviewing the changes and then stumbled upon the following solution(s):
- https://stackoverflow.com/a/33983971/7410631
- https://stackoverflow.com/a/41730436/7410631

Can we use any of these instead? They require adding `order: <number>` to episodes (that don't have numbered prefixes) and an update to `header.html` (to allow both numbered and unnumbered episodes). But it would keep the number of changes we need to make to a minimum.

tobyhodges:
Thank you for pointing me to this in the docs, @brownsarahm - this functionality was only introduced in the most-recent version of Jekyll, released last month, and I hadn't been aware of its existence until now.

I completely agree that it's better to use the core Jekyll functionality rather than introducing new customisation. Unfortunately, GitHub Pages isn't using Jekyll v4.0 (yet) so we can't use this right now. Nevertheless, I don't believe it's worth introducing all this extra customisation into the Carpentries template as a stop-gap - I think instead we should simply wait for the ordering functionality to be available once GH Pages is using Jekyll v. ≥4...

Well, I enjoyed fiddling around under the hood of our lesson template anyway, even if the effort was rendered obsolete in the end 🙃 

I'll close this PR now.

maxim-belkin:
You guys are too fast :)

I saw that "new Jekyll feature" -- it got me excited until I tried it and it didn't work -- that's why it is important to use the same versions of gems as GitHub.

From what I see, [github-pages](https://github.com/github/pages-gem) gem has been a little slow (at least, lately) adopting new Jekyll versions. That's why I think it would be nice to be a little independent of that sluggishness and implement features that we can later "revert" when they are implemented upstream.

I don't oppose to adding this functionality to the carpentries/styles. However, _functionality_ (changes that enable non-numeric episode ordering) and changes to our template (changes that depend on or use that functionality) have to be in separate pull requests. Ideally, the latter would work with our changes and with changes that Jekyll implemented in their new release but that would be a cherry on top.

tobyhodges:
Most of the changes in this PR are to add functionality - it is only the naming in the boilerplate files that changes the actual template i.e. `01-introduction.md` -> `introduction.md` and a reordering of the Extras. I'd be happy to drop those changes and reopen with only the additional Liquid logic allowing custom ordering, if that's something you'd be interested in?

maxim-belkin:
>  I'd be happy to drop those changes and reopen with only the additional Liquid logic allowing custom ordering, if that's something you'd be interested in?

Yes, let's do that.

While you're at it, would you mind doing the following changes?

- make variable names more explicit, e.g. epi ~> episode, prev_url ~> previous_page (or whatever it actually means)

- minimize the amount of changes: 
   * I see that you moved the menu itself to `navbar_section.html`. While this might be a reasonable change (I don't know at the moment), can we do it later? MY understanding is that currently we need to change `for episode in site.episodes` in `navbar.html` to `for episode in ordered_episodes` and create/define that `ordered_episodes` variable somewhere.
  * you create some variables that might not be necessary (e.g. `assign episodes = site.episodes` in `navbar.html`).

tobyhodges:
> Yes, let's do that.

Done

> make variable names more explicit, e.g. `epi` ~> `episode`, `prev_url` ~> `previous_page` (or whatever it actually means)

Good point, thanks. I left some sub-optimal names in `syllabus.html` as the nested for loops prevent me from using the `episode` variable name twice. If you have better suggestions than `epi`, `epislug`, etc I'd be delighted to adopt them.

> I see that you moved the menu itself to `navbar_section.html`. While this might be a reasonable change (I don't know at the moment), can we do it later?

Yes and I've (reluctantly) moved the dropdown construction back to `navbar.html`. However, I am definitely +1 for having a separate `navbar_section.html` file to include collections with arbitrary names. This would allow other projects that use this template to have dropdowns for collections called something other than  "Episodes" and/or "Extras" and I am in favour of flexibility.

> My understanding is that currently we need to change `for episode in site.episodes` in `navbar.html` to `for episode in ordered_episodes` and create/define that `ordered_episodes` variable somewhere.

If you know of a way to construct arrays on-the-fly in Jekyll, please give me a hint and I would be delighted to remove the ugly nested loop solution that I've used so far. Otherwise, we are stuck with the redundancy of looping over all the items twice.

maxim-belkin:
The following should construct a variable called `lesson_episodes`. We can put it in a separate file under the `_includes` folder and source it where necessary. Once we have `lesson_episodes`, we may change `for episode in site.episodes` to `for episode in lesson_episodes`.

```liquid
{% if site.contents %}
    {% assign ordered_episodes_string = "" %}
    {% for episode in site.contents.episodes %}
        {% if ordered_episodes_string != blank %}
            {% ordered_episodes_string = ordered_episodes_string | append: "|" | append: episode %}
        {% else %}
            {% ordered_episodes_string = ordered_episodes_string | append: episode %}
        {% endif %}
    {% endfor %}
    {% assign lesson_episodes = ordered_episodes_string | split: "|" %}
{% else %}
    {% assign lesson_episodes = site.episodes %}
{% endif %}
```

tobyhodges:
Thanks. This would give us a list of the episode slugs but that's not the same as what we get when working with a collection in Jekyll. In fact, it's what we get by accessing `site.content.episodes` in the first place.

Each episode object in `site.episodes` holds the URL, title, etc of the episode, as well as the slug. The `site.contents.episodes` array only contains the slugs, which is why I have been using them to access the appropriate `episode` object in `site.episodes`.

Can you suggest a way to create such an array of objects (rather than strings) on-the-fly?

Additionally, if we can achieve such a thing, I think it would be better to make this `include` accept arguments to prevent it being specific to Episodes i.e. to create a variable such as `ordered_pages` that can then be used to build an ordered list of whatever collection (episodes, extras, ...) we want to work with.

maxim-belkin:
> In fact, it's what we get by accessing `site.content.episodes` in the first place.

Yeah, I realized that on my way back home. Please have a look at #437 where I implemented manual episode ordering and please let me know if you see obvious problems. Changes I make there to the files that are already in the repo are minimal. Manual ordering of "Extras" can be implemented in the same way.

tobyhodges:
`{% assign episode = site.episodes | where: "slug", lesson_episode | first %}`

I did not know about this - very good and exactly what I was looking for. Shall I close this again so you can go ahead with your solution?

maxim-belkin:
> Shall I close this again so you can go ahead with your solution?

Let's keep it open for now.

If you don't mind, I'd like to add you as a co-author on that PR (https://help.github.com/en/articles/creating-a-commit-with-multiple-authors) because you've done a lot of work here.

maxim-belkin:
This has been resolved in e176d3b.

Thank you, Toby, for your excellent work here! 🥇

PR #425, opened 2019-08-21
Blockquote paragraphs lack proper space

kekoziar:
add `blockquote p { padding-bottom: 1em; }` to show paragraph breaks in challenges and call-outs. maintainer discussion indicated that `1em` provided better clarity than `inherit`.

below is a screenshot showing the difference using the **LC Intro to Data** lesson for the example.

blockquote-current-1em

### DIFF
--- a/assets/css/bootstrap.css
+++ b/assets/css/bootstrap.css
@@ -1484,8 +1484,8 @@ blockquote ul:last-child,
 blockquote ol:last-child {
   margin-bottom: 0;
 }
-blockquote * + p { 
-  padding-bottom: 1em; 
+blockquote :not(h2) + p {
+    padding-top: 1em;
 }
 blockquote footer,
 blockquote small,
--- a/assets/css/bootstrap.css
+++ b/assets/css/bootstrap.css
@@ -1484,6 +1484,9 @@ blockquote ul:last-child,
 blockquote ol:last-child {
   margin-bottom: 0;
 }
+blockquote p { 
+  padding-bottom: 1em; 
+}
 blockquote footer,
 blockquote small,
 blockquote .small {
--- a/assets/css/bootstrap.css
+++ b/assets/css/bootstrap.css
@@ -1484,7 +1484,7 @@ blockquote ul:last-child,
 blockquote ol:last-child {
   margin-bottom: 0;
 }
-blockquote p { 
+blockquote * + p { 
   padding-bottom: 1em; 
 }
 blockquote footer,
--- a/assets/css/bootstrap.css
+++ b/assets/css/bootstrap.css
@@ -1484,9 +1484,6 @@ blockquote ul:last-child,
 blockquote ol:last-child {
   margin-bottom: 0;
 }
-blockquote :not(h2) + p {
-    padding-top: 1em;
-}
 blockquote footer,
 blockquote small,
 blockquote .small {
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -148,6 +148,9 @@ font-size: 18px;
 blockquote p {
     margin: 5px;
 }
+blockquote :not(h2) + p {
+    padding-top: 1em;
+}
 
 //----------------------------------------
 // Override Bootstrap settings.


maxim-belkin:
I agree that extra padding adds clarity. Could you please try something like this:

```
blockquote * + p { padding-top: 1em; }
```

to see if we can avoid adding extra padding at the bottom of challenges and call-out boxes.

kekoziar:
`blockquote * + p { padding-top: 1em; }` works for the key points, but not the call-out boxes. 
Screen Shot 2019-08-22 at 8 33 18 AM

kekoziar:
So, I tried adding `padding-bottom: 0;` to `blockquote p:last-child, blockquote ul:last-child, blockquote ol:last-child`

That seemed to do the trick.

Let me know what you think.

Screen Shot 2019-08-22 at 8 37 16 AM

maxim-belkin:
@kekoziar, my apologies for confusing you there. I actually like padding at the bottom but it has to come from the box itself, not from its last element. If you look at your first before-and-after image, you'll see that your changes added spacing at the bottom of the box. So, because you justified your PR as "adding space _between_ paragraphs, I suggested `blockquote * + p`. There is no need to suppress padding at the bottom.

kekoziar:
oops, I misread your suggestion.  Now it adds it to the top.  
Screen Shot 2019-08-22 at 9 32 41 AM

assets/css/bootstrap.css
```diff
@@ -1484,7 +1484,7 @@ blockquote ul:last-child,
 blockquote ol:last-child {
   margin-bottom: 0;
 }
-blockquote p { 
+blockquote * + p { 
   padding-bottom: 1em;
```
maxim-belkin: I think it has to be `padding-top` as we're modding the second element in a pair

maxim-belkin:
`blockquote * + p` selects the first `p` too because `*` matches H2 header.  Could you please change it to `:not(h2)`?

Also, which repo are you checking it against? I just want to make sure that repo is up-to-date with this one.

kekoziar:
I am using a [library carpentry lesson](https://github.com/LibraryCarpentry/lc-data-intro) for the example, since the issue is most obvious in that lesson. Their style is up to date with 9.5.3, but not the subsequent changes. I have been merging the most recent style locally to work on this, which is where the screenshots are from. I am going to merge the updated style in that lesson after we resolved this issue.  

(btw, I noticed this issue with the challenges in the git lesson, but just avoided having two paragraphs, or added a line with only a non-breaking-space to force the extra space. Seeing that this is a problem with other lessons prompted me to look at the style for a solution).

Honestly, you're now suggesting things that are beyond my knowledge of css.  

I don’t understand why the most simple solution is not the best: Adding in bold, 

> blockquote p:last-child,
> blockquote ul:last-child,
> blockquote ol:last-child {
>   margin-bottom: 0;
>   **padding-bottom:0;**
> }

> **blockquote p { 
>         padding-bottom: 1em; 
> }**

It only adds space between paragraphs, modifies only the text, and does not modify the blockquote margins/padding.

maxim-belkin:
```css
blockquote p {
    padding-bottom: 1em;
}
```
adds padding to the bottom of all paragraphs in a `blockquote`, including those `p` that are the last elements in `blockquote`s. As a result -- extra space will appear at the bottom of blockquotes that have `p` as their last element. Selection
```css
blockquote :not(h2) + p {
    padding-top: 1em;
}
```
adds padding to the top of all but the first `p` (the first `p` comes after `h2` so it won't satisfy the selection criterion).

I will have a look at that repo later today...

kekoziar:
> 
> 
> ```css
> blockquote p {
>     padding-bottom: 1em;
> }
> ```
> 
> adds padding to the bottom of all paragraphs in a `blockquote`, including those `p` that are the last elements in `blockquote`s. As a result -- extra space will appear at the bottom of blockquotes that have `p` as their last element. 
> 

but, 
```css
blockquote p:last-child,
blockquote ul:last-child,
blockquote ol:last-child {
   margin-bottom: 0;
   padding-bottom:0;
}
``` 
removes the last space from the last child.

Regardless, what you proposed should work. I'll change it
> ```css
> blockquote :not(h2) + p {
>     padding-top: 1em;
> }
> ```

Here's a screenshot of current vs your proposed change. 
blockquote-current-top-1em

maxim-belkin:
@kekoziar, apologies for the delay -- I needed to look at this a bit closer.

The reason paragraph appear the way they do is due to the fact that in `lesson.scss` we use `blockquote p { margin: 5px; }`. Margins of adjacent HTML elements are not additive -- they are squashed: two elements both with 5px margin will have 5px between them. If you change margin of `blockquote p` in `lesson.scss` to 10 / 12 / 15 px (whatever you think is reasonable... perhaps less than 20px) -- we will achieve the same effect while keeping the number of mods to `bootstrap.scss` to a minimum (which I think is a good idea).

kekoziar:
@maxim-belkin using `em` inherits the size of the font instead of hard-coding a size.  In this way if the font size is changed in future, you don't need to change the hard-coded margin, the margin size is already inherited.

kekoziar:
blockquote-top-1em-not-vs-lesson-scss

maxim-belkin:
Yes, I saw the issue with keypoints in my local tests too... Jekyll, for some reason, inserts paragraphs into list items. Anyways, the first point I was trying to make is that we should probably move your changes to `lesson.scss` (perhaps all previous modifications too).

The second point was that we could update `blockquote p` styling, but as you point out in the above comment it looks like it will introduce more troubles, so let's dismiss it.

kekoziar:
adding
```css
blockquote :not(h2) + p {
    padding-top: 1em;
}
```
to `lesson.scss` doesn't add the issue w/the list items, but adds the same spaces as when in `bootstrap.css`. I think it's a matter of which file do you want it in.

maxim-belkin:
Could you please move changes introduced in this PR to `lesson.scss`?

kekoziar:
I think that did it.

maxim-belkin:
Thank you, Katherine!

fmichonneau:
Thank you both for your work on this! :pray:

PR #428, opened 2019-09-12
use bundler to render lessons

fmichonneau:
@carpentries/lesson-infrastructure-committee please provide feedback on this.

- What do you think of including `bundle update` as part of the command that "make" or "serve" the site? including it reduces the chances of getting error messages for missing or outdated dependencies. It's needed with Docker. For Docker, we could use caching and mount the folder that stores the dependencies (https://github.com/envygeeks/jekyll-docker#usage-3) but this works best if the Maintainers (or lesson contributors) only use Docker other the permissions in the folders get messy and it leads to errors. I don't know if people tend to stick to one approach or mix both. (NB: the Docker image we use for Jekyll has bundler installed by default).

- At least on my system, I have to add `--host 0.0.0.0` to the Docker call otherwise I can't access the rendered site.

- In some contexts, it makes sense to commit the Gemfile.lock to the repo, but in ours, I don't think we want to fix the versions of the gems that GitHub Pages is using, instead, we always want to use the latest versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ _site
 .Rhistory
 .RData
 .vendor/
+.docker-vendor/
 Gemfile.lock
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ _site
 .Rproj.user
 .Rhistory
 .RData
+.vendor/
 Gemfile.lock
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ _site
 .Rproj.user
 .Rhistory
 .RData
-
+Gemfile.lock
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'github-pages', group: :jekyll_plugins
--- a/Gemfile
+++ b/Gemfile
@@ -1,2 +1,10 @@
+# frozen_string_literal: true
+
 source 'https://rubygems.org'
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
+# Synchronize with https://pages.github.com/versions
+ruby '>=2.5.3'
+
 gem 'github-pages', group: :jekyll_plugins
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@
 # Settings
 MAKEFILES=Makefile $(wildcard *.mk)
 JEKYLL_VERSION=3.8.5
-JEKYLL=bundle update && bundle exec jekyll
 JEKYLL_DOCKER=${JEKYLL} serve --host 0.0.0.0
+JEKYLL=bundle install --path .vendor/bundle && bundle update && bundle exec jekyll
 PARSER=bin/markdown_ast.rb
 DST=_site
 
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,9 @@
 
 # Settings
 MAKEFILES=Makefile $(wildcard *.mk)
-JEKYLL=jekyll
 JEKYLL_VERSION=3.8.5
+JEKYLL=bundle update && bundle exec jekyll
+JEKYLL_DOCKER=${JEKYLL} serve --host 0.0.0.0
 PARSER=bin/markdown_ast.rb
 DST=_site
 
@@ -19,7 +20,7 @@ commands :
 
 ## docker-serve     : use docker to build the site
 docker-serve :
-	docker run --rm -it -v ${PWD}:/srv/jekyll -p 127.0.0.1:4000:4000 jekyll/jekyll:${JEKYLL_VERSION} make serve
+	docker run --rm -it -v ${PWD}:/srv/jekyll -p 4000:4000 jekyll/jekyll:${JEKYLL_VERSION} ${JEKYLL_DOCKER}
 
 ## serve            : run a local server.
 serve : lesson-md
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -19,7 +20,7 @@ commands :
 
 ## docker-serve     : use docker to build the site
 docker-serve :
-	docker run --rm -it -v ${PWD}:/srv/jekyll -p 127.0.0.1:4000:4000 jekyll/jekyll:${JEKYLL_VERSION} make serve
+	docker run --rm -it -v ${PWD}:/srv/jekyll -p 4000:4000 jekyll/jekyll:${JEKYLL_VERSION} ${JEKYLL_DOCKER}
```
alee: :wave: @fmichonneau ! This looks like a great start but I had trouble running this on my system, I kept getting these errors:

```
Bundle updated!
/bin/sh: 1: bundle: not found
Makefile:23: recipe for target 'docker-serve' failed
make: *** [docker-serve] Error 127
```

A few other things to note:

1. we may still want to leave the 127.0.0.1 in the docker port mapping so we aren't exposing the jekyll server to the network
2. I think we should install the bundle to `./vendor/bundle` (mapped onto the local filesystem) so we don't have to download/install/update the gems every single time we run the application. 
3. given the increasing complexity of the docker command, it might be good to have a standalone shell script (still very simple) like `jekyll.sh` that does the bundle install, bundle update, and bundle exec jekyll serve.

fmichonneau: Thanks for your work and feedback on this!

Were you getting the error your report using the `latest` jekyll image?

I'll respond to your notes with some comments:

1. In my tests, and from my reading, if we use 127.0.0.1, it doesn't work on Ubuntu by default. Which OS do you use? Do you know if there are settings elsewhere on the system or Docker that need to be changed to make it work? The networking part of Docker isn't something I'm familiar with.

2. I have tried to this, but in my tests, the file permissions interacted in a way that made it impossible to use both the Docker-based approach and the local version of bundler/jekyll. I can imagine that most people who would use Docker will not use the local executables and that might be OK. It'd be great to have more feedback from maintainers who use the Docker-based approach to understand better how they use it. There might also be a way to work around this that I don't know about!

3. That sounds good. We may want to put this script in the `bin/` folder.

alee: I got the error using jekyll 3.8.5 as specified in the Makefile - still not exactly sure why it is happening but I think it has to do with the `&&` chaining which doesn't work as we want it to when run via `docker exec`. I vaguely recall having to run chained commands via `sh -c "foo && bar && baz"` or some other hacky workaround, for which point 3 is a better solution (having a dedicated shell script).

I've got a working example of this at https://github.com/alee/styles/tree/use-bundler and there's a working version of this applied to the python-novice-gapminder lesson at https://github.com/alee/python-novice-gapminder/tree/docker-bundle

You can see the commit where all these changes were made at https://github.com/alee/styles/commit/00245b1bb83e707c2db5bd7a0af16423db693707

1. The docker container definitely needs the --host 0.0.0.0 but here we are binding the docker container's exposed port 4000 to our localhost:4000 so that when we visit localhost:4000 on our host machine, it'll proxy to the docker container's port 4000. I think the issue you may have run into was the missing `jekyll serve --host 0.0.0.0` which I'm guessing is implicit in the normal `jekyll serve` configuration but not the `bundle exec jekyll serve` for some reason..

2. In the code I linked above I run the `bundle install --path .vendor/bundle` so after running `make docker-serve` you should see a new docker managed & created `.vendor/` directory in the root of your repository (which is why there's an additional .gitignore entry there as well). I am not sure where people tend to put there typical bundle install libs but we can pick a different hidden directory as well (like `.docker-vendor/bundle`) that should provide a fairly strong guarantee of not conflicting with their system install

3. Good call!

Let me know if you'd like me to drop those commits into this PR, though some permission adjustment is probably needed to allow me to do that

fmichonneau:
@alee thanks for your help with this! I adopted your suggestions and added you as a co-author of the commit.

Could members of the @carpentries/lesson-infrastructure-committee review this?

.gitignore
```diff
@@ -9,4 +9,6 @@ _site
 .Rproj.user
 .Rhistory
 .RData
-
+.vendor/
+.docker-vendor/
```
maxim-belkin: at what step is `.docker-vendor` created?

fmichonneau: at the `bundle install` step when running `make docker-serve`

maxim-belkin: thank you.

Gemfile
```diff
@@ -0,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'github-pages', group: :jekyll_plugins
```
maxim-belkin: I suggest adding a few extra lines here: 
- `frozen_string_literal` -- make strings unchangeable. Usual practice in Ruby [1](https://stackoverflow.com/a/37799399/7410631).
- `git_source`: workaround for a bug in Bundler [2](https://stackoverflow.com/a/41454572/7410631)
- `ruby '>=2.5.3'` -- to make sure people use an up-to-date Ruby used by GitHub

```
# frozen_string_literal: true

source "https://rubygems.org"
git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }

# Synchronize with https://pages.github.com/versions
ruby '>=2.5.3'

gem 'github-pages', group: :jekyll_plugins
```

fmichonneau:
Thanks Maxim!

PR #436, opened 2019-09-26
restore lost CSS settings

maxim-belkin:
When we updated bootstrap to 3.4.1 in 6e5cba2, we lost a few settings that we had previously made to `bootstrap.css`. In this PR I'm restoring some of those lost settings.
Others either appear to make no changes or make changes that should be discussed.

Changes explained:

* bootstrap.css
  - restore 17.5px font in blockquote. We override it in `lesson.scss`
* lesson.scss
  - override blockquote font-size
  - disallow hyphens in `samp`
  - set top & left margins in `dt` and `dd` elements, correspondingly (affects "References" page)

Settings omitted in this PR:
- `blockquote pre:last-child { margin-botton: 0; }` as it appears to not affect anything
- `code { color: inherit; }` & `pre { color: inherit; }` as combined they appear to change the default text color of code blocks to match that of the language block (e.g. purple for Python)

### DIFF
--- a/assets/css/bootstrap.css
+++ b/assets/css/bootstrap.css
@@ -1476,7 +1476,7 @@ abbr[data-original-title] {
 blockquote {
   padding: 10px 20px;
   margin: 0 0 20px;
-  font-size: inherit;
+  font-size: 17.5px;
   border-left: 5px solid #eeeeee;
 }
 blockquote p:last-child,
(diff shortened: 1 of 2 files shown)


fmichonneau:
thanks for catching this @maxim-belkin and sorry I missed it when I did the upgrade!

maxim-belkin:
No worries, François! I suggest that we keep `bootstrap.css` unchanged going forward and make stylistic changes in `lesson.scss`.

fmichonneau:
yes, I agree!

PR #437, opened 2019-09-27
Manual ordering of episodes and extras

maxim-belkin:
This is an implementation of manual episode ordering. It uses `episode_order` and `extras_order` array variables if they is defined in `_config.yml`. See the comment below: https://github.com/carpentries/styles/pull/437#issuecomment-536710306.

### DIFF
--- a/_includes/aio-script.md
+++ b/_includes/aio-script.md
@@ -1,13 +1,20 @@
 {% comment %}
 As a maintainer, you don't need to edit this file.
-If you notice that something doesn't work, please 
+If you notice that something doesn't work, please
 open an issue: https://github.com/carpentries/styles/issues/new
 {% endcomment %}
+
+{% include manual_episode_order.html %}
 
 <script>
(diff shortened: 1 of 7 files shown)


maxim-belkin:
@fmichonneau, I requested your review of this PR, so here is a brief explanation of what it does. It enables manual ordering of episodes and extras by setting `episode_order` and `extras_order` array variables in `_config.yml` like so:

```yml
episode_order:
 - episodeA
 - episodeB
extras_order:
 - extraA
 - extraB
```

New variables that are created in this PR:

- `lesson_episodes` variable that "replaces" `site.episodes`
- `lesson_extras` variable that "replaces" `site.extras`
-  `previous_episode` variable that replaces `page.previous` in `episode_navbar.html`
- `next_episode` variable that replaces `page.next` in `episode_navbar.html`

fmichonneau:
thanks!

PR #438, opened 2019-11-01
use jekyll to generate all in one instead of Javascript

fmichonneau:
instead of doing Javascript injection to generate the all in one page, I realized we could use a pure Jekyll approach.

I parameterized the "include" commands that generate the objective, questions, keypoints, so we can pass the values of each episode when generating the all in one page.

@carpentries/lesson-infrastructure-committee what do you think of this change?

@maxim-belkin I'd appreciate a double check to make sure it also works with manual episode ordering as I'm not too familiar yet with how this works.

### DIFF
--- a/_includes/aio-script.md
+++ b/_includes/aio-script.md
@@ -1,6 +1,6 @@
 {% comment %}
 As a maintainer, you don't need to edit this file.
-If you notice that something doesn't work, please 
+If you notice that something doesn't work, please
 open an issue: https://github.com/carpentries/styles/issues/new
 {% endcomment %}
 
@@ -21,5 +21,5 @@ open an issue: https://github.com/carpentries/styles/issues/new
 {{ e.content }}
 
 {% include episode_keypoints.html episode_keypoints=e.keypoints %}
-<hr/>
+<hr />
 {% endfor %}
--- a/_includes/aio-script.md
+++ b/_includes/aio-script.md
@@ -1,50 +1,25 @@
 {% comment %}
 As a maintainer, you don't need to edit this file.
-If you notice that something doesn't work, please
+If you notice that something doesn't work, please 
 open an issue: https://github.com/carpentries/styles/issues/new
 {% endcomment %}
 
 {% include manual_episode_order.html %}
 
-<script>
-  window.onload = function() {
-    var lesson_episodes = [
-    {% for lesson_episode in lesson_episodes %}
-      {% if site.episode_order %}
-        {% assign episode = site.episodes | where: "slug", lesson_episode | first %}
-      {% else %}
-        {% assign episode = lesson_episode %}
-      {% endif %}
-    "{{ episode.url}}"{% unless forloop.last %},{% endunless %}
-    {% endfor %}
-    ];
-    var xmlHttp = [];  /* Required since we are going to query every episode. */
-    for (i=0; i < lesson_episodes.length; i++) {
-      xmlHttp[i] = new XMLHttpRequest();
-      xmlHttp[i].episode = lesson_episodes[i];  /* To enable use this later. */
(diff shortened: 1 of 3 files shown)


maxim-belkin:
1. Going away from Javascript is definitely a good thing: the burden of generating this page will be on GitHub, not on viewer's machine. Moreover, it will be generated by GitHub only once and not every time that all-in-one page is visited, which is the case with the Javascript version.

2. This PR works well with manual episode ordering. To test it, I pulled the PR to my local repository of "Python Novice Inflammation" lesson and added the following to my `_config.yml`:
```yaml
episode_order:
   - 06-func
   - 01-numpy
```
And generated all-in-one page had the correct order of episodes.

A few comments:
1. It makes sense to move `aio-script.md` from the `_includes` directory (not sure where) and rename it to `all-in-one.md`.
2. We should consider using Liquid's `{%- -%}` instead of `{% %}` to not produce empty lines in generated HTML files. [link](https://shopify.github.io/liquid/basics/whitespace/)
3. Kramdown by default adds ids to the header elements, so no need to do it manually.

_includes/aio-script.md
```diff
@@ -1,50 +1,25 @@
 {% comment %}
 As a maintainer, you don't need to edit this file.
-If you notice that something doesn't work, please
+If you notice that something doesn't work, please
```
maxim-belkin: 🤨

_includes/aio-script.md
```diff
@@ -1,50 +1,25 @@
...
-  {% endif %}
-  <article id="{{ episode.url }}"></article>
+<h1 id="{{ e.title | slugify }}" class="maintitle">{{ e.title }}</h1>
+
+{% include episode_overview.html teaching_time=e.teaching exercise_time=e.exercises episode_questions=e.questions episode_objectives=e.objectives %}
+
+{{ e.content }}
+
+{% include episode_keypoints.html episode_keypoints=e.keypoints %}
+<hr/>
```
maxim-belkin: Do we need this `hr`? If so then, technically, there has to be a space before /

_includes/episode_keypoints.html
```diff
@@ -1,10 +1,17 @@
 {% comment %}
   Display key points for an episode.
 {% endcomment %}
+
+{% if page.keypoints == null %}
```
maxim-belkin: I'm a tad confused here as "nothing" in Liquid is `nil`, not `null`. Also, we could check whether we have something in `page.keypoints` with `{%- if page.keypoints -%}` and avoid having to check `== null` completely. The only case where`page.keypoints == nil` is if `page.keypoints` can be an empty string, which evaluates to "true" in Liquid.

brownsarahm:
I also think that moving away from js is a great idea for the reasons mentioned.

Additionally, the fewer different technologies that the whole setup relies on, the easier it is for new community members to get up to speed on how the pieces work together. 

It would be nice to add some comments so that it's a bit easier for someone who wants to understand what it's doing to figure that out.  In particular, why the ifs are there might not be obvious to someone who has only worked with the regular episode pages, not the AIO page.

maxim-belkin:
François, is anything blocking this PR? Can I somehow help?

fmichonneau:
help would be wonderful, the changes you and Sarah suggested sound good to me. If you can update the PR with them it'd be happy to review and merge.

naupaka:
I am +1 for Jekyll over Javascript, based not on my expertise but on the principle of keeping the rendering in one place.

maxim-belkin:
> help would be wonderful, the changes you and Sarah suggested sound good to me. If you can update the PR with them it'd be happy to review and merge.

OK, will do.

fmichonneau:
thanks!

PR #440, opened 2019-11-06
Switch to Liquid comments

maxim-belkin:
HTML comments end up in the generated HTML pages: they're not displayed by
the browsers but they're still present there. Liquid comments do not end up
in the generated HTML pages


maxim-belkin:
🎉

Issue #445, opened 2019-12-02
Building/Setting up lesson on Windows fails.


JoaoRodrigues:
Hi there,

I'm trying to setup a new lesson on Windows 10 and I'm running into several issues that prevent me from generating the lesson website.

I am not running Git Bash nor WSL. I have both Ruby and Jekyll installed, and I installed make for Windows from the GnuWin32 page. I imported the styles repo (https://github.com/joaorodrigues/intro-git-github) and cloned it locally. Here's a list of steps and the issues that I found (with some workarounds):

1. Running `python bin/lesson_initialize.py` works and copies all the files:
```
_config.yml  _episodes_rmd/  _includes/  aio.md   AUTHORS  CITATION  CODE_OF_CONDUCT.md  data/  files/    LICENSE.md  README.md     setup.md
_episodes/   _extras/        _layouts/   assets/  bin/     code/     CONTRIBUTING.md     fig/   index.md  Makefile    reference.md
```

2. Running `make lesson-check` fails with a Python error. The scripts define `/usr/bin/env python3`, they should define `python` since not everyone has a python3 alias.

3. Running `make lesson-check` still fails, this time with an encoding problem:
```
index.md:6:FIXME: home page introduction
setup.md:4:FIXME
_episodes/01-introduction.md:6:- "Key question (FIXME)"
_episodes/01-introduction.md:8:- "First learning objective. (FIXME)"
_episodes/01-introduction.md:10:- "First key point. Brief Answer to questions. (FIXME)"
_episodes/01-introduction.md:12:FIXME
reference.md:7:FIXME
_extras/discuss.md:4:FIXME
_extras/guide.md:4:FIXME
Exception in thread Thread-5:
Traceback (most recent call last):
  File "C:\Users\joaor\Miniconda3\envs\scw\lib\threading.py", line 917, in _bootstrap_inner
    self.run()
  File "C:\Users\joaor\Miniconda3\envs\scw\lib\threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\joaor\Miniconda3\envs\scw\lib\subprocess.py", line 1238, in _readerthread
    buffer.append(fh.read())
  File "C:\Users\joaor\Miniconda3\envs\scw\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 8243: character maps to <undefined>

[...]

PR #446, opened 2019-12-03
Python scripts: compatibility with Windows

JoaoRodrigues:
This PR includes several commits that improve compatibility of the code on windows machines. For a detailed description of the problems and fixes, see #445 .


bin/lesson_check.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
maxim-belkin: `python` still refers to Python 2 on most platforms. Please revert this and all similar changes.

bin/lesson_check.py
```diff
@@ -117,7 +117,6 @@ def main():
     check_config(args.reporter, args.source_dir)
     check_source_rmd(args.reporter, args.source_dir, args.parser)
     args.references = read_references(args.reporter, args.reference_path)
-
```
maxim-belkin: Please undo this change

bin/lesson_check.py
```diff
@@ -29,19 +29,19 @@
...
+    'README.md': False,
+    os.path.join('_extras', 'discuss.md'): True,
+    os.path.join('_extras', 'guide.md'): True,
+    'index.md': True,
+    'reference.md': True,
+    'setup.md': True,
 }
 
 # Episode filename pattern.
-P_EPISODE_FILENAME = re.compile(r'/_episodes/(\d\d)-[-\w]+.md$')
```
maxim-belkin: How do you handle removal of `/_episodes/` here?

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
+        dpath = os.path.join(source_dir, d)
+
+        pattern = os.path.join(dpath, '*.md')
+        for filename in glob.glob(pattern):
+            data = read_markdown(parser, filename)
+            if data:
+                result[filename] = data
+
+    return result
+
```
maxim-belkin: This is an unconditional `return`, so the rest of the function becomes unreachable after this line

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
     result = {}
     for pat in all_patterns:
+        print(pat)
         for filename in glob.glob(pat):
             data = read_markdown(parser, filename)
             if data:
+                print(filename)
                 result[filename] = data
-    return result
+#    return result
```
maxim-belkin: why?

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
+        pattern = os.path.join(dpath, '*.md')
+        for filename in glob.glob(pattern):
+            data = read_markdown(parser, filename)
+            if data:
+                result[filename] = data
+
+    return result
+
     all_dirs = [os.path.join(source_dir, d) for d in SOURCE_DIRS]
+    print(all_dirs)
```
maxim-belkin: please remove this line

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
     all_patterns = [os.path.join(d, '*.md') for d in all_dirs]
+    print(all_patterns)
```
maxim-belkin: please remove this line

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
     result = {}
     for pat in all_patterns:
+        print(pat)
```
maxim-belkin: please remove this line

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
         for filename in glob.glob(pat):
             data = read_markdown(parser, filename)
             if data:
+                print(filename)
```
maxim-belkin: please remove this line

bin/lesson_check.py
```diff
@@ -257,37 +256,56 @@ def read_all_markdown(source_dir, parser):
...
+    required = [os.path.join(source_dir, p) for p in REQUIRED_FILES]
     missing = set(required) - set(filenames_present)
     for m in missing:
         reporter.add(None, 'Missing required file {0}', m)
 
     # Check episode files' names.
     seen = []
-    for filename in filenames_present:
-        if '_episodes' not in filename:
+    for filepath in filenames_present:
```
maxim-belkin: please undo the unnecessary `filename` -> `filepath` change

bin/util.py
```diff
@@ -118,6 +118,7 @@ def read_markdown(parser, path):
     cmd = 'ruby {0}'.format(parser)
     p = Popen(cmd, shell=True, stdin=PIPE, stdout=PIPE,
               close_fds=True, universal_newlines=True)
+    body = str(body.encode('utf-8'))
```
maxim-belkin: Can you explain why this change is necessary? Under what circumstances `body` has different encoding?

bin/lesson_check.py
JoaoRodrigues: Python 2 will not be supported from 2020. Reverting this to `python3` breaks this code in my system because I don't have such an alias. Since I am using a fresh install of Anaconda3, I assume others will have the same problem.

The correct path forward, if you wish to continue supporting Python 2, is to write a function that checks the version explicitly. I could do that if you prefer.

bin/lesson_check.py
JoaoRodrigues: The removal of the folder itself? We'd have to check for it explicitly. The variable is called 'filename' so I assumed it would suffice to check for the filename itself, not the directory name.

bin/lesson_check.py
JoaoRodrigues: Sorry, this is actually a leftover from what I was working on. The code below is the old implementation. I'll remove it.

bin/lesson_check.py
JoaoRodrigues: I changed the name of the variable to reflect the fact that it contains more than just the file name. I split it below using `os.path.basename` and I assign the result of that to `fname`.

bin/util.py
JoaoRodrigues: I explained why this is happening in #445 and why this is necessary to fix the code to run on (any) windows (10, maybe others) machines. I agree it's not elegant, but it's the quickest patch I could think of that should not break in other OSes.

bin/lesson_check.py
maxim-belkin: > The correct path forward, if you wish to continue supporting Python 2, is to write a function that checks the version explicitly. I could do that if you prefer.

This script does not work with Python 2.

bin/lesson_check.py
maxim-belkin: You removed `/_episodes/` from the pattern. How do you handle this removal?

bin/lesson_check.py
maxim-belkin: let's keep changes to a minimum here. We can rename variables in a separate PR.

bin/util.py
maxim-belkin: there is no explanation in #445.

bin/lesson_check.py
JoaoRodrigues: In `check_fileset`, I split the path explicitly. I only check the regular expression against the basename of the path (ie, the name of the file itself).

```
    # Check episode files' names.
    seen = []
    for filepath in filenames_present:
        if '_episodes' not in filepath:
            continue

        # split path to check episode name
        fname = os.path.basename(filepath)
        m = P_EPISODE_FILENAME.search(fname)
        if m and m.group(1):
            seen.append(m.group(1))
        else:
            reporter.add(
                None, 'Episode {0} has badly-formatted filename', filepath)
```

bin/lesson_check.py
maxim-belkin: This change would break thee script in Linux (and, hence, Travis and other CI platforms) and macOS.

JoaoRodrigues: Perfect, then the more reason to add an explicit check at the beginning to check for that, instead of relying on an alias that is not present in all architectures/OSes. Using `python3` does not work on Windows 10 with a fresh installation of Anaconda 3. These are the executables I have in the PATH:

```
C:\Users\SoftwareCarpentry\Work\sc-teaching\intro-git-github $ python
python.exe     pythontex.exe  pythonw.exe
```

bin/lesson_check.py
JoaoRodrigues: Done.

bin/util.py
JoaoRodrigues: Copied from there:

Then I get the encoding problem. From this [StackOverflow answer](https://stackoverflow.com/a/33291200), this seems to be a by-product of using `universal_newlines=True`. This has also been reported as a bug in Python as well (report [here](https://bugs.python.org/issue27179)). We can get around it by specifying the encoding directly, which is what I did previously, by changing line 121.

bin/lesson_check.py
maxim-belkin: Explicit check would not prevent the failure you're experiencing. 
Shebang would fail before the check can be implemented. We can replace the shebang with a note to use Python3 which might be `python` or `python3` depending on platform.

bin/lesson_check.py
maxim-belkin: thank you

bin/lesson_check.py
JoaoRodrigues: The problem is that the setup makes use of `make` to call these python scripts and does so by executing them like this `./script.py`, which triggers the shebang. If I run the script manually (`python script.py`) indeed, I don't have this problem.

Maybe a better workaround is to modify the `Makefile` to allow people to edit which python is being used to call the scripts.

bin/lesson_check.py
JoaoRodrigues: What would be your suggestion for variable names then?

bin/lesson_check.py
maxim-belkin: We can detect which Python executable to use (python or python3) in a Makefile or in a new script that would then call this script.

maxim-belkin: In general, I'm leaning towards removing the shebang rather than changing it (changing it might cause more troubles)

JoaoRodrigues: Agreed. On a Mac, or on Linux, what interpreter shows up when you type `python`?

bin/lesson_check.py
```diff
@@ -257,37 +257,42 @@ def read_all_markdown(source_dir, parser):
...
-        for filename in glob.glob(pat):
+    for d in SOURCE_DIRS:
+        dpath = os.path.join(source_dir, d)
+
+        pattern = os.path.join(dpath, '*.md')
+        for filename in glob.glob(pattern):
             data = read_markdown(parser, filename)
             if data:
                 result[filename] = data
+
```
maxim-belkin: this change looks good but it not related to Win10 compatibility. Let's do this in a different PR.

bin/lesson_check.py
maxim-belkin: keep `filename`. `fname` can be `bname` or `basename` in this PR. Again, we can change variable names in a separate PR.

bin/lesson_check.py
maxim-belkin: Mac:
```sh
$ /usr/bin/python -V
Python 2.7.16
```

brownsarahm: On Ubuntu 16.04:
```
$ /usr/bin/python -V
Python 2.7.12
```

Even though I set my path with Anaconda so that the following works:
```
$ python
Python 3.6.4 |Anaconda, Inc.| (default, Jan 16 2018, 18:10:19) 
[GCC 7.2.0] on linux
```

bin/lesson_check.py
JoaoRodrigues: Done.

bin/util.py
maxim-belkin: Can you try replacing `Popen` + `p.communicate` with `subprocess.run(cmd, capture_output=True)`?

bin/lesson_check.py
maxim-belkin: thank you!

bin/util.py
JoaoRodrigues: This works: `p = subprocess.run(cmd, capture_output=True, input=body, text=True, encoding='utf-8')`

Which is just a wrapper for what we had before :)

JoaoRodrigues: Without the `encoding` parameter, it fails.

bin/lesson_check.py
JoaoRodrigues: I deleted the shebang lines and added a `PYTHON` variable to the Makefile. It's defined as `python`, but it should be easier to modify to whatever executable we want. Does this work as a solution? See the latest commit to this PR.

Makefile
```diff
@@ -93,15 +94,15 @@ _episodes/%.md: _episodes_rmd/%.Rmd
...
 
 ## lesson-check-all : validate lesson Markdown, checking line lengths and trailing whitespace.
 lesson-check-all :
-	@bin/lesson_check.py -s . -p ${PARSER} -r _includes/links.md -l -w --permissive
+	${PYTHON} bin/lesson_check.py -s . -p ${PARSER} -r _includes/links.md -l -w --permissive
 
 ## unittest         : run unit tests on checking tools.
 unittest :
-	@bin/test_lesson_check.py
+	${PYTHON} bin/test_lesson_check.py
```
maxim-belkin: As is, this change does not aid Win 10 compatibility. Let's improve Makefile in a separate PR.
(Also, let's keep those `@`)

bin/util.py
maxim-belkin: I see. Let's add `encoding='utf-8'` to the `Popen` call instead of `body = str(body.encode('utf-8'))`.
(btw, why do you need `str(...)`?)

JoaoRodrigues: Done. 

(`str.encode` returns a byte-string. `str()` converts to a regular text string)

Makefile
JoaoRodrigues: See my general comment on the PR.

JoaoRodrigues:
Hi @maxim-belkin,

I got a bit lost with all the comments and diverging conversations on the PR so I'm summarizing here.

This PR addresses compatibility with Windows systems (at least for Win 10, which I can test) by fixing 3 issues:
1. Hard-coded paths in the `lesson-check.py` file, in particular in regular expressions.
2. Encoding issues when parsing/reading files within the `lesson-check.py` script following an external process (subprocess.Popen).
3. The absence of `python3` aliases or executables in some distributions/versions of Python on Windows, namely those recommended by the Carpentries (Anaconda3).

I believe this PR, as is, addresses all three points coherently. The change in the `Makefile` does not fix the problem, but provides an extremely trivial solution. All that is left to do is add a comment in the setup instructions for Windows users to change the `PYTHON` variable in the `Makefile`.

As for the other changes that might be considered cosmetic, to be completely honest, I had an incredibly hard time understanding (at a glance) what some methods were doing in this code. If I changed things, it was because I had to rewrite parts of the code to understand _how_ to fix it. I believe the result is more readable and more maintainable.

As such, I am not willing to work on this PR further. You're welcome to fork my repo and cherry-pick the changes and commit them yourself to the main branch/repo. I spent a more than a few hours figuring out what the problem was and how to fix it, then working on a PR, and writing an issue that rivals chapters of my PhD thesis in length. 

The lessons are up and running on my laptop, that's all I wanted. Fixing it for everybody else was a bonus. You're all volunteers with limited time to devote to this project, but so am I. I hope you understand my point of view.

Makefile
JoaoRodrigues: (Put the @s back in)

bin/util.py
maxim-belkin: > (`str.encode` returns a byte-string. `str()` converts to a regular text string)

oh, right. I usually `.decode()` these.

maxim-belkin:
Thanks for your contribution, @JoaoRodrigues. I'll take the PR from here, so please don't delete your fork and the branch you used to submit the PR from (win10_compatibility).

Regarding your last comment: conversations are not diverging, but there are quite a few of them, which makes things hard to follow. The usual recommendation (and a common practice) is to submit smaller pull requests. They are easier to discuss and usually take less time and effort on both sides (contributors' and maintainers').

maxim-belkin:
I think I caught all Windows incompatibilities. 
Handing off to you, @fmichonneau and @brownsarahm.

maxim-belkin:
I'm going to test this on a Windows machine.

hoytpr:
I am anxious to test this on my machine, but got lost in the code (sorry).  I see seven files changed... my Github repo "data-cleanup" is essentially a styles repo import. Would it make sense to import/merge the changes from  JoaoRodrigues:win10_compatibility there, then re-clone locally to test?

maxim-belkin:
in the `data-cleanup` directory, execute:
```sh
curl -LO https://github.com/carpentries/styles/pull/446.patch # if you have `curl`
# or
wget https://github.com/carpentries/styles/pull/446.patch # if you have wget
git apply 446.patch
```
and you should have changes in this PR in your lesson directory.

hoytpr:
Thanks @maxim-belkin does it matter that your patch files have permissions 644 when mine were 755?

maxim-belkin:
it shouldn't matter: Makefile has been updated to call `python some-script.py`. Previously, Makefile was instructing Make to execute these scripts -- this is why they had executable bits and shebangs (shebang is the `#!/usr/bin/env ...` line in the script)

hoytpr:
Apparently it does matter. I can't run any of the scripts from the patch. Some researching sees similar issues, but no fixes because of the way Windows assigns permissions. I'm going to push them up to my Github repo, and pull them back down.

brownsarahm:
you will definitely need execute permissions. 

This appears to be a way to correct permissions on windows in gitbash:
https://medium.com/@akash1233/change-file-permissions-when-working-with-git-repos-on-windows-ea22e34d5cee

hoytpr:
In a meeting now. Will try again ~3:30pm

maxim-belkin:
The article you linked to talks about the opposite issue: files created on Windows lack executable bits necessary on Linux machines to "execute" scripts (that is do `./script.sh` instead of `bash script.sh`).

Here we no longer execute scripts themselves but execute Python which reads the scripts and executes their contents. I'm currently testing this PR on Windows and I can call Python to execute scripts `python bin/lesson_initialize.py` without any error messages even though there are no executable bits.

It looks like there is a big problem with Gnu Make 3.81 (that João is/was using) -- http://gnu-make.2324884.n4.nabble.com/process-begin-CreateProcess-NULL-quot-quot-failed-td13819.html

This is a very old problem and the solution is to build Make from source while applying 2 patches or install a newer version of Gnu make (from ezwinports or somewhere else).

maxim-belkin:
> Apparently it does matter. I can't run any of the scripts from the patch.

Could you please elaborate a little bit more? What do you mean by "scripts from the patch". Patch contains changes to the scripts, not the scripts themselves. One has to apply the patch with `git apply 446.patch`.

---
Regarding Gnu Make:

although it is possible to install Gnu Make that we need (the one from ezwinports works fine), I think we have to mark GnuWin32 as `unsupported` -- the developer stopped supporting it 2 years ago and `Make` that it ships with has a bug that is important for us.

maxim-belkin:
@hoytpr, could you please describe the tools you use (GitBash, Miniconda, etc...) when you get a chance? I'd like to test this PR using them.

hoytpr:
@maxim-belkin 

By this time I had already run "python bin/lesson_initialize.py"
You sent me the `curl` link, and I downloaded the patch into my "data-cleanup" repo.
I entered `git apply 446.patch` and got many warnings similar to:
`warning: bin/lesson_check.py has type 100644, expected 100755`
but no "errors" or "failures"
Then entered `make lesson-check` and got the error
```
make: execvp: python3: Permission denied
make: *** [Makefile:112: lesson-check] Error 127
```
When I viewed my repo with `ls -lart` I remember seeing new python script files in the `/bin` directory. I tried to run the lesson_check.py directly but the response was:
```
$ python lesson_check.py   ## note underscore: this was "lesson_check" not "lesson-check"
Path to Markdown parser not provided
````
Then, being new to this, I tried to re-apply the patch, and there were many errors including:
```
$ git apply 446.patch 
error: patch failed: bin/util.py:118
error: bin/util.py: patch does not apply
warning: bin/lesson_check.py has type 100644, expected 100755
error: patch failed: bin/lesson_check.py:1
error: bin/lesson_check.py: patch does not apply
warning: bin/lesson_initialize.py has type 100644, expected 100755
```
I'm going to delete the repo, and re-clone it from Github 
-----------------------
My machine:

I'm using a Win10 machine and GitBash. When I type:
```
$ bash --version
GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```
I have Anaconda3 (64-bit) installed, although when I open the Anaconda (or GitBash) terminal and type:
```
(base) C:\Users\Hoyt>anaconda --version
anaconda Command line client (version 1.7.2)
```
Others:
```
$ python --version
Python 3.7.1
$ bundle --version
Bundler version 2.0.2
$ ruby --version
ruby 2.5.7p206 (2019-10-01 revision 67816) [x64-mingw32]
$ jekyll --version
jekyll 4.0.0
$ git --version
git version 2.21.0.windows.1

R version 3.6.1 (2019-07-05) -- "Action of the Toes"
Copyright (C) 2019 The R Foundation for Statistical Computing
Platform: x86_64-w64-mingw32/x64 (64-bit)
```
Through Anaconda Navigator I have:
RStudio 1.1.456
Qt Console 4.4.3
jupyter Notebook 5.7.4
JupyterLab 0.35.3

**Comment by maxim-belkin** [comment]
> By this time I had already run "python bin/lesson_initialize.py"
> You sent me the `curl` link, and I downloaded the patch into my "data-cleanup" repo.
> I entered `git apply 446.patch` and got many warnings similar to:
> `warning: bin/lesson_check.py has type 100644, expected 100755`
> but no "errors" or "failures"

Warning are normal in this case because the `curl` command that I provided downloads the entire PR as a single patch file and one of the commits in this patch removes executable permission bits from python scripts.


> Then entered `make lesson-check` and got the error
> 
> ```
> make: execvp: python3: Permission denied
> make: *** [Makefile:112: lesson-check] Error 127
> ```

This is strange: it shouldn't list `python3` but `python` because you're on Windows. Unless your shell has `python3` command... what does `which python3` show?

> When I viewed my repo with `ls -lart` I remember seeing new python script files in the `/bin` directory. I tried to run the lesson_check.py directly but the response was:
> 
> ```
> $ python lesson_check.py   ## note underscore: this was "lesson_check" not "lesson-check"
> Path to Markdown parser not provided
> ```

If you look into the `Makefile,` you'll see a line:

```
@${PYTHON} bin/lesson_check.py -s . -p ${PARSER} -r _includes/links.md
```

And `PARSER` is `bin/markdown_ast.rb`. If you look at that Ruby file, you'll see that it uses two Gems: `json` and `kramdown`, which you need to install with `gem`:

```sh
gem install json kramdown
```

After that you should be able to execute lesson checks manually like this (GitBash):

```sh
# cd to the root of your lesson repository
python bin/lesson_check.py -s . -p bin/markdown_ast.rb -r _includes/links.md
```

hoytpr:
@maxim-belkin 
> This is strange: it shouldn't list python3 but python because you're on Windows. Unless your shell has python3 command... what does which python3 show?
```
$ which python3                                     
/c/Users/Hoyt/AppData/Local/Microsoft/WindowsApps/python3
```
```
$ gem install json kramdown
Temporarily enhancing PATH for MSYS/MINGW...
Building native extensions. This could take a while...
Successfully installed json-2.2.0
Parsing documentation for json-2.2.0
Done installing documentation for json after 1 seconds
Successfully installed kramdown-2.1.0
Parsing documentation for kramdown-2.1.0
Done installing documentation for kramdown after 1 seconds
2 gems installed
```
```
$ python bin/lesson_check.py -s . -p bin/markdown_ast.rb -r _includes/links.md
Traceback (most recent call last):
  File "bin/lesson_check.py", line 566, in <module>
    main()
  File "bin/lesson_check.py", line 115, in main
    check_config(args.reporter, args.source_dir)
  File "bin/lesson_check.py", line 176, in check_config
    config = load_yaml(config_file)
  File "C:\Users\Hoyt\git\Saved\data-cleanup\bin\util.py", line 164, in load_yaml
    return yaml.load(reader, Loader=yaml.FullLoader)
AttributeError: module 'yaml' has no attribute 'FullLoader'
```

brownsarahm:
the method it says is missing is new, via (https://stackoverflow.com/questions/55551191/module-yaml-has-no-attribute-fullloader), that issue shows how to check  and update it if you use pip.  There is a way to do it with conda too, if you prefer

since @hoytpr also has a python3 alias, maybe his install of anaconda was older than Joao's and anaconda changed how it sets the alias?

maxim-belkin:
> ```
> $ python bin/lesson_check.py -s . -p bin/markdown_ast.rb -r _includes/links.md
> Traceback (most recent call last):
>   File "bin/lesson_check.py", line 566, in <module>
>     main()
>   File "bin/lesson_check.py", line 115, in main
>     check_config(args.reporter, args.source_dir)
>   File "bin/lesson_check.py", line 176, in check_config
>     config = load_yaml(config_file)
>   File "C:\Users\Hoyt\git\Saved\data-cleanup\bin\util.py", line 164, in load_yaml
>     return yaml.load(reader, Loader=yaml.FullLoader)
> AttributeError: module 'yaml' has no attribute 'FullLoader'
> ```

As Sarah explained, this error messages is caused by the outdated PyYAML module.
Because you're using Anaconda, I would suggest using `conda` command to update Python packages, though the end result should not depend on what tool you use. So, run

```sh
conda update
```

You might also want to clean up unused packages after this:

```sh
conda clean --all
```



> since @hoytpr also has a python3 alias, maybe his install of anaconda was older than Joao's and anaconda changed how it sets the alias?

I think you're right.
Pete, try to update `conda` and Anaconda itself with:

```sh
conda update conda
conda update anaconda
```
You might need to restart GitBash after that. See if you still have `python3`. But note that even if you do -- it shouldn't be a problem and shouldn't cause any.

JoaoRodrigues:
Chiming in to add info on this issue of the python call. This answer on SO seems to explain why there is not python3.exe on Windows and provides an alternative: https://stackoverflow.com/a/34223417

In addition, would you be interested in setting up a CI like appveyor or Travis to check if the make commands pass on Linux, Win, and MacOS? I would be happy to work on that to improve testing/debugging these issues.

hoytpr:
FYI: The alias appears to have been installed as part of a Windows update for "Zune" (including iHeartRadio) on Nov 5, 2019. 
I went ahead and ran:
```
$ Python  
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> yaml.__version__
'3.13'
quit()

$ conda update PyYAML
Solving environment: done

## Package Plan ##
  environment location: C:\Users\Hoyt\Anaconda3
  added / updated specs:
    - pyyaml
The following packages will be downloaded:
    package                    |            build
    ---------------------------|-----------------
    conda-4.7.12               |           py37_0         3.0 MB
    conda-package-handling-1.3.11|           py37_0         280 KB
    ------------------------------------------------------------
                                           Total:         3.3 MB

The following NEW packages will be INSTALLED:
    conda-package-handling: 1.3.11-py37_0
The following packages will be UPDATED:
    conda:                  4.5.12-py37_0 --> 4.7.12-py37_0
Proceed ([y]/n)? y
```
**Because this looked to only update PyYaml to 4.7, I ran `conda update PyYAML` again**
```
$ conda update PyYAML
Collecting package metadata (repodata.json): done
Solving environment: /
The environment is inconsistent, please check the package plan carefully
The following packages are causing the inconsistency: 
  - defaults/win-64::anaconda==2018.12=py37_0
  - defaults/win-64::blaze==0.11.3=py37_0
  - defaults/win-64::numba==0.41.0=py37hf9181ef_0
```
Went ahead with resolving the inconsistencies, and the next thing I know I have 299
files being updated or installed, plus:
```
The following packages will be DOWNGRADED:

  anaconda                                   2018.12-py37_0 --> custom-py37_1
  jupyter_console                              6.0.0-py37_0 --> 5.2.0-py37_1
```

Went ahead with this and then:

```
$ Python
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> yaml.__version__
'5.1.2'
>>>
```

hoytpr:
Little advice please. 
I ran `conda clean --all` and it removed a bunch of older files, but then it says:
```
...(etc)
Removed zstd-1.3.7-h508b16e_0.tar.bz2
Removed _anaconda_depends-2019.03-py37_0.tar.bz2
Removed _ipyw_jlab_nb_ext_conf-0.1.0-py37_0.tar.bz2
WARNING: C:\Users\Hoyt\.conda\pkgs does not exist
Cache location: C:\Users\Hoyt\Anaconda3\pkgs
Will remove the following packages:
C:\Users\Hoyt\Anaconda3\pkgs
----------------------------

anaconda-2018.12-py37_0                       59 KB
anaconda-custom-py37_1                         6 KB
anaconda-project-0.8.2-py37_0                2.8 MB
asn1crypto-0.24.0-py37_0                     867 KB
astroid-2.1.0-py37_0                         1.3 MB
astropy-3.1-py37he774522_0                  28.7 MB
(etc.)...
Proceed ([y]/n)?
```

Should I continue?
-p

maxim-belkin:
These are outdated packages, so it should be safe to remove them.

hoytpr:
Okay,  this is where I am now;

```
$ python
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> yaml.__version__
'5.1.2'
quit()
$ python bin/lesson_check.py -s . -p bin/markdown_ast.rb -r _includes/links.md
.\README.md: Internally-defined links may be missing definitions: "issues"=>"FIXME"
```

maxim-belkin:
> Okay, this is where I am now;
> 
> ```
> $ python
> Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import yaml
> >>> yaml.__version__
> '5.1.2'
> quit()
> $ python bin/lesson_check.py -s . -p bin/markdown_ast.rb -r _includes/links.md
> .\README.md: Internally-defined links may be missing definitions: "issues"=>"FIXME"
> ```

This is a success test. The error message means that `README.md` has a link that is not defined, e.g. `[text][issues]`. To fix this, add a line to the bottom of the file:

```md
issues: https://some.location
```

hoytpr:
The page rendered using `bundle exec jekyll serve`:

First-render

hoytpr:
Looks good. Adding
```
issues: https://some.location
``` 
to the bottom of `README.md` fixed the error during rendering. 
```
$ bundle exec jekyll serve
C:/Ruby25-x64/lib/ruby/2.5.0/x64-mingw32/etc.so: warning: already initialized constant Struct::Passwd
Configuration file: C:/Users/Hoyt/git/Saved/data-cleanup/_config.yml
            Source: C:/Users/Hoyt/git/Saved/data-cleanup
       Destination: C:/Users/Hoyt/git/Saved/data-cleanup/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 1.32 seconds.
 Auto-regeneration: enabled for 'C:/Users/Hoyt/git/Saved/data-cleanup'
    Server address: http://127.0.0.1:4000
  Server running... press ctrl-c to stop.
```

-p

maxim-belkin:
> Looks good. Adding
> 
> ```
> issues: https://some.location
> ```
> 
> to the bottom of `README.md` fixed the error during rendering.

This is good news. Basically, you were able to detect and fix an issue in `README.md` using the `lesson_check.py` script directly. Does `make lesson-check` work? What does `make --version` report?

hoytpr:
```
$ make --version
GNU Make 4.2.1
Built for i686-pc-msys
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
I still get an error about python3:
```
$ make lesson-check
index.md:6:FIXME: home page introduction
index.md:14:> FIXME
setup.md:4:FIXME
_episodes/01-introduction.md:6:- "Key question (FIXME)"
_episodes/01-introduction.md:8:- "First learning objective. (FIXME)"
_episodes/01-introduction.md:10:- "First key point. Brief Answer to questions. (FIXME)"
_episodes/01-introduction.md:12:FIXME
reference.md:7:FIXME
_extras/discuss.md:4:FIXME
_extras/guide.md:4:FIXME
make: execvp: python3: Permission denied
make: *** [Makefile:112: lesson-check] Error 127
```

hoytpr:
```
$ which python
/c/Users/Hoyt/Anaconda3/python
$ which python3
/c/Users/Hoyt/AppData/Local/Microsoft/WindowsApps/python3
```

hoytpr:
Possible solution here:
https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10

maxim-belkin:
Thanks, Pete. Yes, that `python3` is a link to Microsoft Store to install Python 3. I'll have to work around this in the logic (in the Makefile) that determines which Python to use.

hoytpr:
I turned off the aliases using the SO site suggestion, and re-ran `make lesson-check` getting only "Error 1":

```
$ make lesson-check
/usr/bin/which: no python3 in (/c/Users/Hoyt/bin:/mingw64/bin:/usr/local/bin:/usr/bin:MY_WHOLE_PATH_WAS_DISPLAYED_HERE)
index.md:6:FIXME: home page introduction
index.md:14:> FIXME
setup.md:4:FIXME
_episodes/01-introduction.md:6:- "Key question (FIXME)"
_episodes/01-introduction.md:8:- "First learning objective. (FIXME)"
_episodes/01-introduction.md:10:- "First key point. Brief Answer to questions. (FIXME)"
_episodes/01-introduction.md:12:FIXME
reference.md:7:FIXME
_extras/discuss.md:4:FIXME
_extras/guide.md:4:FIXME
.\README.md: Internally-defined links may be missing definitions: "issues"=>"FIXME"
make: *** [Makefile:112: lesson-check] Error 1
```

maxim-belkin:
`Error 1` is what we want -- this is the code that `lesson-check` raises when encountering an issue.

hoytpr:
So disabling the Windows aliases does work, although when the makefile asks `which python3` it echos the entire PATH line to the screen. But, I am worried if we ask learners to disable those aliases,  Windows will keep turning them back on after updates. I can test this as I have an update to install.

maxim-belkin:
I've added the logic that should deal with Python3 from the MS Store. Could you please test it, Pete? You should be able to "restore" your repository to the state before applying the patch with:

```sh
git checkout -- .
rm 446.patch
```

After that, you should be able to re-download the patch with `curl` and apply it again.

hoytpr:
With a caveat it seemed to work:
CAVEAT: I ran
```
$ bin/lesson_initialize.py
```
without prepending the command with `python` and got:
```
bin/lesson_initialize.py: line 1: Initialize a newly-created repository.: command not found
bin/lesson_initialize.py: line 4: import: command not found
bin/lesson_initialize.py: line 5: import: command not found
bin/lesson_initialize.py: line 6: import: command not found
bin/lesson_initialize.py: line 8: syntax error near unexpected token `('
bin/lesson_initialize.py: line 8: `BOILERPLATE = ('
```
So I then ran:
``` 
$ python bin/lesson_initialize.py
```
And got no response at all. I assume this is good.
```
$ ls                                                                                  
 _config.yml    _includes  assets    code                fig         Makefile
_episodes      _layouts   AUTHORS   CODE_OF_CONDUCT.md  files       README.md
_episodes_rmd  446.patch  bin       CONTRIBUTING.md     index.md    reference.md
_extras        aio.md     CITATION  data                LICENSE.md  setup.md
```
So I then ran:
```
$ make lesson-check
index.md:6:FIXME: home page introduction
index.md:14:> FIXME
setup.md:4:FIXME
_episodes/01-introduction.md:6:- "Key question (FIXME)"
_episodes/01-introduction.md:8:- "First learning objective. (FIXME)"
_episodes/01-introduction.md:10:- "First key point. Brief Answer to questions. (FIXME)"
_episodes/01-introduction.md:12:FIXME
reference.md:7:FIXME
_extras/discuss.md:4:FIXME
_extras/guide.md:4:FIXME
.\README.md: Internally-defined links may be missing definitions: "issues"=>"FIXME"
make: *** [Makefile:120: lesson-check] Error 1
```
Looks good!

hoytpr:
Rendered fine.
```
$ bundle exec jekyll serve 
C:/Ruby25-x64/lib/ruby/2.5.0/x64-mingw32/etc.so
Configuration file: C:/Users/Hoyt/git/Saved/data-cleanup/_config.yml
            Source: C:/Users/Hoyt/git/Saved/data-cleanup
       Destination: C:/Users/Hoyt/git/Saved/data-cleanup/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 1.996 seconds.
 Auto-regeneration: enabled for 'C:/Users/Hoyt/git/Saved/data-cleanup'
    Server address: http://127.0.0.1:4000
  Server running... press ctrl-c to stop.
``
![second-render](https://user-images.githubusercontent.com/5910884/70482066-2114df80-1aaa-11ea-99a1-f391ac6ed246.png)

**Comment by maxim-belkin** [comment]
> With a caveat it seemed to work:
> CAVEAT: I ran
> 
> ```
> $ bin/lesson_initialize.py
> ```
> 
> without prepending the command with `python` and got:
> 
> ```
> bin/lesson_initialize.py: line 1: Initialize a newly-created repository.: command not found
> bin/lesson_initialize.py: line 4: import: command not found
> bin/lesson_initialize.py: line 5: import: command not found
> bin/lesson_initialize.py: line 6: import: command not found
> bin/lesson_initialize.py: line 8: syntax error near unexpected token `('
> bin/lesson_initialize.py: line 8: `BOILERPLATE = ('
> ```

This is expected: when you applied the patch, Git informed you that it expected Python scripts to not have executable permission bits but found that your scripts still have them. Git applied the patch anyway but did not change their executable permission bits -- that's why you still can execute the scripts (e.g.,  `./bin/lesson_check.py`). When this PR is merged, these scripts will lose their executable permission bits so people won't be able to execute them directly. You can see the effect of it by removing executable permission bits from the scripts manually:
```sh
chmod 0644 bin/*.py   # or:   chmod a-x bin/*.py
```
You will no longer be able to execute the scripts and Git Bash will tell you something like this:
```sh
-bash: ./bin/lesson_check.py: Permission denied
```



> So I then ran:
> 
> ```
> $ python bin/lesson_initialize.py
> ```
> 
> And got no response at all. I assume this is good.

Yes, this is good. Notifying a user of a successful initialization is a reasonable thing to do but this is out of scope of this PR.


> Rendered fine.

Wonderful!

PR #451, opened 2019-12-21
Simpler and more robust Makefile

maxim-belkin:
Reworked the Makefile to make it simpler (to understand) and more robust so that people with no experience with Make would see clear error messages when something is not working.

Clear error messages require things like: 
```Makefile
target:
ifneq (, $(TOOL))
    @$(TOOL) subcommand
else
    $(error sorry can't do)
endif
```

To avoid repetition, I extracted these `ifneq .... endif` to `commands.mk` and created `TOOL_CMD` for each `TOOL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,73 +1,51 @@
 # Use /bin/bash instead of /bin/sh
 export SHELL = /bin/bash
 
-## ========================================
-## Commands for both workshop and lesson websites.
+# #################################################
+#              ** Customizations **
 
-# Settings
-MAKEFILES=Makefile $(wildcard *.mk)
-JEKYLL_VERSION=3.8.5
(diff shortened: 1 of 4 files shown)


maxim-belkin:
@fmichonneau, did you want to discuss this PR or did you want someone else to look at it?

fmichonneau:
Overall, I like that this approach leads to more informative error messages. I do worry about the increasing complexity of the file. It would also make things easier to manage if everything was in a file, is it possible? Finally, we need to make sure that we can run `make serve` and `make site` without requiring an internet connection (it seems that the curl call would prevent this).

maxim-belkin:
I'll rebase this PR once #473 is merged and update it as requested so that we can discuss it again.

Issue #454, opened 2020-02-05
warnings rendering as errors from rmd files


LunaSare:
Hello! I am using the workshop-template to develop a lesson on how to use a couple R packages. To show how functions works, I am using rmarkdown.
It is working great, except that warnings from functions are rendered as errors in the website, here https://lunasare.github.io/ssb2020_workshop/02-get-induced/index.html, rmd code here https://github.com/LunaSare/ssb2020_workshop/blob/gh-pages/_episodes_rmd/02-get-induced.Rmd, md code here https://github.com/LunaSare/ssb2020_workshop/blob/gh-pages/_episodes/02-get-induced.md

I could modify the md file, but I would have to do that after every time I render. Is there a way to specify this on the rmd file?

Thanks!

brownsarahm:
Thanks for the issue @LunaSare. I moved it to the styles repository (which is the template for new lessons, like you're creating) from the workshop website template that's used to create the ad pages for workshops (which are the two day events).  These terms are tricky, but you should get help here.  

It looks like at least part of the issue is that there is not a css style defined to syntax highlight warnings differently that errors per: https://github.com/carpentries/styles/blob/gh-pages/assets/css/lesson.scss

@maxim-belkin can confirm/ guide how to fix that, because I don't know how the R code running and producting the error/warning text is tagged.

maxim-belkin:
Thank you, Sarah.

Luna: Thank you for the report. What would you like to do with the output you're currently getting? Would you like to see the section to be called "Warning" and not "Error"? Or did you envision something else?

Sarah is right in that we don't currently have a separate style for warnings. We'd have to create one before the word "Warning" can appear. It seems like a reasonable thing to have but I'd like to discuss it more before implementing it.

If you'd like to do something else... chances are, you'd have to do it manually at this point. But then, again, please let us know what you'd like to achieve and we'll be happy to help.

François ( @fmichonneau ), I tried adding `class.warning = 'output'` to the code block in question and it produced:
```html
... class="language-{.output}" ...
```
so it looks like `chunk-options.R` needs to be updated slightly.

fmichonneau:
Thanks for bringing this up @LunaSare! 
We currently don't have support for warnings in our template :grimacing: so it'd be a great thing to have. I proposed a possible solution to have it in #455, once it's merged, I can put in a PR on your repository to bring the changes in.

fmichonneau:
closed by #455

LunaSare:
[...]

PR #455, opened 2020-02-06
add warning hook + CSS class for Rmd-based lessons

fmichonneau:
proposed implementation for #454 

Preview:
Screenshot from 2020-02-06 10-31-23

@carpentries/lesson-infrastructure-committee what do you think? I'm happy to consider a different color scheme :art:


maxim-belkin:
I haven't noticed `warning = hook_error` in `chunk-options.R` when I looked at it. Nice work!

PR #458, opened 2020-03-18
update LICENSE and COC links

zkamvar:
Since the user is always directed here, the links should point to the correct pages. The COC link is a bit bare, however since the relevant content is Jekyll-derived. Maybe we should just link the Carpentries code of conduct instead?

### DIFF
--- a/bin/boilerplate/CONTRIBUTING.md
+++ b/bin/boilerplate/CONTRIBUTING.md
@@ -10,12 +10,12 @@ and reviews of proposed changes are all welcome.
 ## Contributor Agreement
 
 By contributing,
-you agree that we may redistribute your work under [our license](LICENSE.md).
+you agree that we may redistribute your work under [our license](../../LICENSE.md).
 In exchange,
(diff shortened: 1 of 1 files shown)


maneesha:
This came up in another context before, where GitHub automatically links to a project's COC file.  I agree that this should link to [our CoC](https://docs.carpentries.org/topic_folders/policies/index_coc.html).

maxim-belkin:
Thank you for the PR, Zhian.
Thank you for your comment, Maneesha.

When maintainers synchronize lessons with `carpentries/styles` repository, they copy files from `bin/boilerplate` directory to the corresponding lessons' directories. So, `bin/boilerplate/CONTRIBUTING.md` is copied to the root folder replacing the current `CONTRIBUTING.md` file. The change suggested in this PR would, therefore, brake the link.

We can link to https://docs.carpentries.org/topic_folders/policies/index_coc.html, however, I feel that we should discuss this change before implementing it

PR #461, opened 2020-05-27
Use carpentries/lesson-docker for docker-serve make rule

zkamvar:
This adds the https://github.com/carpentries/lesson-docker container to the styles repository, overwriting the jekyll docker container.

The `docker-serve` make rule originally relied on the Jekyll docker container, which posed some problems with development regarding permissions. In my testing, I found that there was a mismatch between my user (1001) and the user Jekyll assumed (1000), which threw an error that it couldn't write Gemfile.lock

Moreover, maintainers/contributors still needed R on their machines if they wanted to develop Rmarkdown lessons.

Finally, the version of Jekyll was fixed in the makefile and did not update along with GitHub. 

This PR should address these issues and also does the following:

1. adds .*history to the .gitignore to avoid any history files from being accidentally written
2. removes the $JEKYLL_VERSION variable since that's taken care of by the lesson-docker repository


fmichonneau:
Thank you!!!

PR #464, opened 2020-06-19
lesson.scss: style tab panels on setup pages

maxim-belkin:
Improve styling of tab panels on setup pages:

Before
image

After:

image


fmichonneau:
Thank you! that was a much needed change :sparkles:

PR #468, opened 2020-06-26
OS stripe color

maxim-belkin:
I propose to tweak OS stripe background color:

image

I also noticed that links don't have good-enough contrast even on F9F9F9 color -- the color used in tables on Carpentries lessons' websites -- so I propose to tweak that one as well. Contrast ratios with the new color are excellent: above 7 on most backgrounds.

I used https://webaim.org/resources/contrastchecker to check contrast ratio


fmichonneau:
thanks Maxim!

maxim-belkin:
Thanks, François!

PR #469, opened 2020-06-29
Upgrade jQuery to 3.5.1

maxim-belkin:
Title says it all.

Note: we can't use '.slim' version of jQuery that lacks `ajax` and `effects` modules because we use `.toggle()` and `.hide()` methods provided by the `effects` module.


fmichonneau:
thanks Maxim!

tomgreen66:
Looks like `.size()` is deprecated in the new version of jQuery and instead use `.length`.  This currently breaks solutions. See: 
https://github.com/carpentries/styles/blob/5d4aaa2fb38e78d161af6f84ed0fde3abed5c8b9/assets/js/lesson.js#L7-L8

maxim-belkin:
Thanks for the report, Thomas! Fixed in 3001ab5

PR #472, opened 2020-07-03
License is not a copyright. License info ID

maxim-belkin:
Copyright and license are different things. So I propose to rename the class of the `div` element containing the license information from `copyright` to `license` and add an `id` attribute so that we can refer to it in the header using:

```
<link rel="license" href="#license-info" />
```

This is the way to declare which license applies to the page. Alternatively, we can also point directly to `href="{{ site.cc_by_human }}` as we do in the footer.

### DIFF
--- a/_includes/lesson_footer.html
+++ b/_includes/lesson_footer.html
@@ -6,7 +6,7 @@
 
 <footer>
   <div class="row">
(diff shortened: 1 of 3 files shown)


fmichonneau:
thanks Maxim!

PR #473, opened 2020-07-10
Deprecated use of --path when installing bundle

ocaisa:


Makefile
```diff
@@ -6,7 +6,7 @@ export SHELL = /bin/bash
 
 # Settings
 MAKEFILES=Makefile $(wildcard *.mk)
-JEKYLL=bundle install --path .vendor/bundle && bundle update && bundle exec jekyll
+JEKYLL=bundle config set path '.vendor/bundle' && bundle install && bundle update && bundle exec jekyll
```
maxim-belkin:
```
bundle config set path '.vendor/bundle'
```
sets this configuration globally (that is, in `~/.bundle/config` file). Let's add `--local` flag so that it reads:

```
bundle config --local set path .vendor/bundle
```

maxim-belkin: Also, we might not need single quotes around `.vendor/bundle`

maxim-belkin (approved):
Thank you, @ocaisa! Suggested changes look good to me!

fmichonneau:
thank you both!

PR #482, opened 2020-08-04
repo_check.py: allow URLs not ending with .git

maxim-belkin:
New pattern allows `repo_check.py` script to handle cases when URL
is something like `https://github.com/carpentries/lesson-name`.


fmichonneau:
thanks Maxim, that sounds reasonable.

PR #483, opened 2020-08-05
repo_check.py: match https repositories

maxim-belkin:
Current pattern matches `git@github.com:/USER/repo` URLs but doesn't `https://github.com/USER/repo`.
This PR allows https repositories


fmichonneau:
thanks!

PR #484, opened 2020-08-05
Makefile: don't use /bin/bash shell

maxim-belkin:
On some platforms `bash` executable is in `/bin` folder, whereas on other system itis in `/usr/bin`.
Instead of assuming that it's in the `/bin` directory, we can check whether `/bin/bash` or `/usr/bin/bash` exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,3 @@
-# Use /bin/bash instead of /bin/sh
-ifneq (,$(wildcard /bin/bash))
-  export SHELL = /bin/bash
-else ifneq (,$(wildcard /usr/bin/bash))
-  export SHELL = /usr/bin/bash
-endif
-
 ## ========================================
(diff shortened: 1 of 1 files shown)


fmichonneau:
I don't know a lot about this topic, but I thought that using `#!/usr/bin/env bash` was a safer approach.

maxim-belkin:
Git for Windows doesn't have /usr/bin/env so checking which file exists is easier.

maxim-belkin:
Alternatively, we can use `$(shell which bash)` People use /usr/bin/env ... in shebangs mostly. When one needs to determine the location of an executable in a script, `which ...` is used.

maxim-belkin:
Let's actually, remove this line -- I don't think we use the specifics of Bash shell in our Makefile

fmichonneau:
thanks Maxim!

PR #485, opened 2020-08-07
Fix Python scripts for Windows: UTF-8 encoding

maxim-belkin:
the default encoding on Windows is cp1252, so
we have to specify that the files we open have a different (UTF-8) encoding.


maxim-belkin:
This is ready now.

fmichonneau:
Thank you for fixing this!

PR #487, opened 2020-08-07
GitHub Actions: check labels

maxim-belkin:
GitHub Action to check labels in the repository every Monday and on every push to `gh-pages` branch.

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -1,0 +1,30 @@
+name: Labels
+on:
+  schedule:
+    - cron: '0 12/24 * * 1' #  ~> every Monday at 12:00 UTC
+  push:
+    branches: gh-pages
+jobs:
+  check-labels:
+    name: Check repository labels
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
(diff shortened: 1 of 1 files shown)


fmichonneau:
maybe we could use something like this: https://github.com/micnncim/action-label-syncer ?

maxim-belkin:
It'd be great to use an action for that but there is one unknown with this 'action-label-syncer' action: how does it handle changes to labels' colors and names? There is an (unanswered) issue with that question -- https://github.com/micnncim/action-label-syncer/issues/44. Seeing the timeline on open issues/PRs, I'm a bit skeptical about using it but we can do that.

Regardless of whether we use this action or not, we need to decide how to handle labels:

1. Enforce labels according to a file in a lesson repository.
2. Enforce labels according to a file in some other (`carpentries/style`?) repository.

With (1) above, labels can get out of sync (unless we create a scheduled action to sync lessons with carpentries/styles)
With (2) above, we need to set up Personal Access Tokens for each lesson organization (SWC, DC, LC. Someone with admin privileges would have to do it).

And I also suggest we set up default labels for all four orgs (carpentries, swcarpentry, datacarpentry, librarycarpentry):
https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/managing-default-labels-for-repositories-in-your-organization

PR #488, opened 2020-08-07
GitHub Actions: website

maxim-belkin:
GitHub Action to check that lesson website builds fine and passes `make lesson-check` and `make lesson-check-all` tests.

This action can be amended to push to `gh-pages` branch for lessons written in R-markdown.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -1,0 +1,55 @@
+name: Website
+on:
+  push:
+    branches: gh-pages
+  pull_request: []
+jobs:
+  build-website:
+    name: Build lesson website
+    if: github.repository != 'carpentries/styles'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Set up Ruby
+        uses: actions/setup-ruby@main
+        with:
+          ruby-version: '2.7.1'
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install GitHub Pages, Bundler, and kramdown gems
+        run: |
+          gem install github-pages bundler kramdown
+
+      - name: Install Python modules
+        run: |
+          python3 -m pip install --upgrade pip setuptools wheel pyyaml==5.3.1 requests
+
+      - name: Checkout the lesson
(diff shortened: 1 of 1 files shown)


.github/workflows/website.yml
```diff
@@ -0,0 +1,55 @@
+name: Website
+on:
+  push:
+    branches: gh-pages
+  pull_request: []
+jobs:
+  build-website:
+    name: Build lesson website
+    if: github.repository != 'carpentries/styles'
```
fmichonneau, line 8: I'm wondering if it'd be worth to change the logic here, so that on `carpentries/styles` we create an extra step to run `python ./bin/lesson_initialize.py` before running the `make lesson-check`. There might be a few other things to do manually on a newly initialized lesson before it passes the checks.

.github/workflows/website.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Look for R-markdown files
+        id: check-rmd
+        run: |
+          echo "::set-output name=count::$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})"
+
+      - name: Install R packages
```
fmichonneau: we probably need to import the same steps as in the other PR (#489) for R-based lessons.

maxim-belkin: Yes, logic has to be the same here and there. I'll update it once we decide on what we need here in that other PR.

.github/workflows/website.yml
fmichonneau: that would give us an additional check in this repo to make sure that lessons can be initialized without any issues (and ideally that the newly initialized lesson passes the checks).

maxim-belkin: Wouldn't it be better to do this as a separate GH Action "New Lesson Setup" or as a second job in "template.yml"?

.github/workflows/website.yml
maxim-belkin: I imported the steps for lessons in Rmarkdown from `template.yml` into here.

fmichonneau:
thanks!

PR #489, opened 2020-08-07
GitHub Actions: check lesson template

maxim-belkin:
Action to dry-run changes we're making here to this repository on a number of lessons. I chose:
- swcarpentry/shell-novice
- datacarpentry/r-intro-geospatial
- librarycarpentry/lc-git

Tests are performed on all three operating systems: macos, linux, and windows (git for bash)

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           r-version: 'release'
 
-      - name: Install neeeded packages
+      - name: Install needed packages
         if: steps.check-rmd.outputs.count != 0
         run: |
           install.packages(c('remotes', 'rprojroot', 'renv', 'desc'))
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -92,7 +92,7 @@ jobs:
             eval $cmd
           done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "18.04"), sep = "\n")')
 
-      - name: Install lesson package dependencies
+      - name: Install R lessons package dependencies
         if: steps.check-rmd.outputs.count != 0
         working-directory: lesson
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/template.yml
```diff
@@ -0,0 +1,96 @@
...
+          install.packages('remotes', 'rprojroot', 'renv')
+        shell: Rscript {0}
+
+      - name: Cache R packages
+        if: runner.os != 'Windows' && steps.check-rmd.outputs.count != 0
+        uses: actions/cache@v1
+        with:
+          path: ${{ env.R_LIBS_USER }}
+          key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
+          restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
```
maxim-belkin: @fmichonneau, I believe we have to swap the order of "Cache R packages" and "Install R packages" and check for cache hits (`steps.cache-r.outputs.cache-hit != 'true'`).

fmichonneau: these are the packages that are needed for all the R-based lessons. The idea is that these packages would always be installed fresh but the packages that are lesson-specific get cached.

maxim-belkin: Currently `Cache R packages` step is not used -- the idea is to skip package installation step when `steps.cache-r.outputs.cache-hit == 'true`.

Don't we also need 'knitr', 'stringr', 'checkpoint', 'ggplot2'?

fmichonneau: not anymore, we now use a combination of `renv` + `rprojroot` which takes care of identifying all the dependencies in our lessons that are then installed as part of the build process.

maxim-belkin: > not anymore, we now use a combination of `renv` + `rprojroot` which takes care of identifying all the dependencies in our lessons that are then installed as part of the build process.

Great!

So, right now the caching step is used to cache downloads only and that's exactly what you want. I'm 👍 with that.

maxim-belkin: wait, I think current order has to be switched, though -- "Install R packages" step won't see any cache because cache is created only after "Install R packages" is executed. Is there a way to download R packages before installing them?

maxim-belkin: we can have `cache-restore` and `cache-save` steps before and after we install R packages.

maxim-belkin: Resolved in https://github.com/maxim-belkin/styles/pull/3

PR #490, opened 2020-08-08
OS stripe: adjust line height

maxim-belkin:
Adjust line height of links in the OS stripe on workshop-template website
so that clickable area is at least 44px high.


fmichonneau:
thanks!

PR #491, opened 2020-08-11
only display Episodes drop-down if we have episodes to show

fmichonneau:
follow up from: https://github.com/carpentries/carpentries-theme/issues/6#issuecomment-671973655

Workshop overview sites don't have episodes so hiding the Episodes in this context makes sense.


maxim-belkin (approved):
👍 makes sense.

PR #494, opened 2020-08-17
Sync styles first

fmichonneau:
@maxim-belkin I think we need to do the sync of the styles repo earlier so the latest changes are available before we attempt to build the lessons.

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -46,6 +46,17 @@ jobs:
           path: lesson
           fetch-depth: 0
 
+      - name: Sync lesson with carpentries/styles
+        working-directory: lesson
+        run: |
+          git config --global user.email "team@carpentries.org"
+          git config --global user.name "The Carpentries Bot"
+          git remote add styles https://github.com/carpentries/styles.git
+          git config --local remote.styles.tagOpt --no-tags
+          git fetch styles ${{ steps.styles-ref.outputs.ref }}:styles-ref
+          git merge -s recursive -Xtheirs --no-commit styles-ref
+          git commit -m "Sync lesson with carpentries/styles"
+
       - name: Look for R-markdown files
         id: check-rmd
         working-directory: lesson
@@ -108,17 +119,6 @@ jobs:
           else
             echo "::set-output name=ref::gh-pages"
           fi
-
-      - name: Sync lesson with carpentries/styles
-        working-directory: lesson
-        run: |
-          git config --global user.email "team@carpentries.org"
-          git config --global user.name "The Carpentries Bot"
-          git remote add styles https://github.com/carpentries/styles.git
-          git config --local remote.styles.tagOpt --no-tags
-          git fetch styles ${{ steps.styles-ref.outputs.ref }}:styles-ref
-          git merge -s recursive -Xtheirs --no-commit styles-ref
-          git commit -m "Sync lesson with carpentries/styles"
 
       - run: make site
         working-directory: lesson
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -81,7 +81,7 @@ jobs:
       - name: Install needed packages
         if: steps.check-rmd.outputs.count != 0
         run: |
-          install.packages(c('remotes', 'rprojroot', 'renv', 'desc'))
+          install.packages(c('remotes', 'rprojroot', 'renv', 'desc', 'rmarkdown', 'knitr'))
         shell: Rscript {0}
 
       - name: Query dependencies
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -46,6 +46,15 @@ jobs:
           path: lesson
           fetch-depth: 0
 
+      - name: Determine the proper reference to use
+        id: styles-ref
+        run: |
+          if [[ -n "${{ github.event.pull_request.number }}" ]]; then
+            echo "::set-output name=ref::refs/pull/${{ github.event.pull_request.number }}/head"
+          else
+            echo "::set-output name=ref::gh-pages"
+          fi
+
       - name: Sync lesson with carpentries/styles
         working-directory: lesson
         run: |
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -121,12 +121,12 @@ HTML_DST = \
 
 ## * install-rmd-deps : Install R packages dependencies to build the RMarkdown lesson
 install-rmd-deps:
-	Rscript -e 'source("bin/dependencies.R"); install_dependencies(identify_dependencies())'
+	Rscript -e 'source(file.path("bin", "dependencies.R")); install_dependencies(identify_dependencies())'
```
maxim-belkin: `bin` and `dependencies.R` lose double quotations marks around them in Windows. Perhaps, we need to save this into a new `.R` file in the bin directory.

Makefile
```diff
@@ -121,12 +121,12 @@ HTML_DST = \
...
 
 ## * lesson-md        : convert Rmarkdown files to markdown
-lesson-md : ${RMD_DST}
+lesson-md    : ${RMD_DST}
```
maxim-belkin:
```suggestion
lesson-md : ${RMD_DST}
```

.github/workflows/template.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     defaults:
       run:
         shell: bash # forces 'Git for Windows' on Windows
+    env:
+      RSPM: 'https://packagemanager.rstudio.com/cran/__linux__/bionic/latest'
```
maxim-belkin, line 20: Does this let us use precompiled binaries?

.github/workflows/template.yml
```diff
@@ -89,36 +111,8 @@ jobs:
         run: |
           while read -r cmd
           do
-            eval $cmd
+            eval sudo $cmd
```
maxim-belkin, line 114: Why is `sudo` necessary -- we're `root` in GitHub Actions.

maxim-belkin:
> I think we need to do the sync of the styles repo earlier so the latest changes are available before we attempt to build the lessons.

You beat me to it :) Yes, we need to sync lessons with this repo before we do R-stuff, so other than a couple of minor comments -- this looks good to me.

fmichonneau:
Thanks for the comments Maxim. I ran out of time today to finish this, but I'll continue working on it on Wednesday.

.github/workflows/template.yml
fmichonneau: yes, so it installs faster

.github/workflows/template.yml
fmichonneau: not sure, but without sudo it doesn't work :(
https://github.com/carpentries/styles/runs/992384878#step:14:12

fmichonneau:
@zkamvar could you please review the changes I made to the R scripts that take care of identifying and installing the package dependencies?

Working on this is making me consider whether we should have R libraries per lessons (that we could manage with `renv`), so we don't touch people's personal libraries...

Makefile
```diff
@@ -121,12 +121,12 @@ HTML_DST = \
 
 ## * install-rmd-deps : Install R packages dependencies to build the RMarkdown lesson
 install-rmd-deps:
-	Rscript -e 'source("bin/dependencies.R"); install_dependencies(identify_dependencies())'
+	@bin/install_r_deps.sh
```
maxim-belkin:
```suggestion
	@${SHELL} bin/install_r_deps.sh
```

I suggest avoid relying on executable bit + shebang in the script -- we might run into issues on Windows platforms. Instead, we can use shell to execute the script.

bin/install_r_deps.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+
```
maxim-belkin:
```suggestion

```
we don't use any of the Bash features here.

bin/install_r_deps.sh
```diff
@@ -0,0 +1,4 @@
...
+Rscript -e "source(file.path('bin', 'dependencies.R')); install_required_packages(); install_dependencies(identify_dependencies())"
```
zkamvar, line 1: Instead of using a bash file to call a multi-line RScript as a single string, why not just make this into an R file and call it with Rscript in the make file? This way it will be easier to read.

```r
source(file.path('bin', 'dependencies.R'))
install_required_packages()
install_dependencies(identify_dependencies())
```

maxim-belkin: François had it not in the script first but it was not behaving correctly on Windows, see: https://github.com/carpentries/styles/runs/994865525?check_suite_focus=true#step:15:11

zkamvar:
> @zkamvar could you please review the changes I made to the R scripts that take care of identifying and installing the package dependencies?
> 
> Working on this is making me consider whether we should have R libraries per lessons (that we could manage with `renv`), so we don't touch people's personal libraries...

The script looks a lot cleaner and I think it's a good step going forward :+1: 

Regarding using {renv}, I think it's a good idea! From what I can tell, {renv} is MUCH MUCH kinder than {packrat} was. When you initialize {renv}, you get a lot of text explaining what's happening and there are a lot more helpful user features. I just tested it out on my mac, and it does appear to not interfere with user packages. Since {renv} now has the capability of restoring from a MRAN snapshot, issues of unavailable binaries should be no problem.

The only issue that I can forsee is that the user will need to install ALL the packages of the repo up front and we will want to figure out a way of providing clear updates to people who want to update the renv.lock file.

maxim-belkin:
> The only issue that I can forsee is that the user will need to install ALL the packages of the repo up front and we will want to figure out a way of providing clear updates to people who want to update the renv.lock file.

Once this is figured out, we can save that knowledge as a new target in the Makefile, but this shouldn't be a blocker for this PR.

zkamvar:
> > The only issue that I can forsee is that the user will need to install ALL the packages of the repo up front and we will want to figure out a way of providing clear updates to people who want to update the renv.lock file.
> 
> Once this is figured out, we can save that knowledge as a new target in the Makefile, but this shouldn't be a blocker for this PR.

To clarify: I am just talking about the _very first_ time a user or maintainer interacts with this project. Once they have the {renv} cache set up, it will know where to pull files from and the process should be seamless. There should be no need for makefile intervention on that front.

maxim-belkin (approved):

fmichonneau:
Thank you Maxim and Zhian. I'll open a separate issue about using {renv}

PR #496, opened 2020-08-19
fix: tighten definitions of highlighter

henryiii:
The current style sheet is too greedy in trying to apply highlighter styles; when you use a structure like this:

```
{% highlight cmake %}
{% include code/00-intro/CMakeLists.txt %}
{% endhighlight %}
```

Then you end up with different structure, where the highlighter styles get applied to something that is not a div, and you get something like this:

This tightens up the definitions a little to make sure they don't get applied to non-divs.

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -38,44 +38,44 @@ $color-testimonial: #fc8dc1 !default;
     border-radius: 4px 0 0 4px;
 }
 
-.error   { @include cdSetup($color-error); }
-.warning { @include cdSetup($color-warning); }
-.output  { @include cdSetup($color-output); }
-.source  { @include cdSetup($color-source); }
+div.error   { @include cdSetup($color-error); }
+div.warning { @include cdSetup($color-warning); }
+div.output  { @include cdSetup($color-output); }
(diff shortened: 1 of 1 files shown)


fmichonneau:
this looks good to me. Could someone else double-check?

maxim-belkin:
I think classes without `language-` prefix would need `div` prefixed as well for consistency. Otherwise, LGTM as well.

henryiii:
I did that in my original patch, but left it off here since I think those are manually applied, rather than auto-generated; but if they were somehow applied to something that was not a div, it would probably still break similarly. Will add back into this PR, then.

maxim-belkin (approved):
Tested locally on one of the lessons -- everything works as before, so 👍

fmichonneau:
thank you both!

maxim-belkin:
> I did that in my original patch, but left it off here since I think those are manually applied, rather than auto-generated

Honestly, I'm not sure why these `.bash`, `.python`, etc are still there -- `{: .xxx}` where `xxx` is not prefixed with `language-` do nothing (at least, nothing one may expect them to be doing) and `{% highlight xxx %}` are not affected by our mods.

PR #497, opened 2020-08-19
add three more common languages

henryiii:
These are common languages, and are needed (for example) for the C++ and CMake trainings being worked on for the HSF.

(C, C++, and CMake)


maxim-belkin (approved):
No objections from me! Looks great!
Thanks, Henry!

fmichonneau:
thank you!

PR #500, opened 2020-08-24
GitHub Actions: better workflow and job names

maxim-belkin:
This is an attempt to improve the workflow and job names. Open for suggestions/discussion.
Currently, job names are too long and are cut off:

image

so one has to click "Details" to go to the Actions tab to see what was tested where. I suggest to shorten job names a little bit so that we can see all the important info without having to click anywhere:

image

I'm not sure if putting `()` around the Carpentry names improves things so suggestions are welcome.

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -1,11 +1,11 @@
-name: Template
+name: Test template
 on:
   push:
     branches: gh-pages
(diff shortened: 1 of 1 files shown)


fmichonneau:
that's a good improvement, thanks Maxim!

PR #501, opened 2020-08-29
GitHub Action: website.yml: don't run in forks

maxim-belkin:
Our "Build website" Action runs in forks. I propose to not run it in people's forks and restrict it to the 4 Carpentries' organizations (not sure we need 'carpentries' though). Unfortunately, proposed syntax is the only syntax that I could come up with to achieve this goal (and I'm not sure there is a better solution).

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -5,7 +5,7 @@ on:
   pull_request: []
 jobs:
   build-website:
-    if: github.repository != 'carpentries/styles'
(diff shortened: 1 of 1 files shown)


brownsarahm:
We do need 'carpentries', for instructor training. 

This would mean, though, that if someone wants to serve a fork of a lesson they would have to modify this file?  

For example, I've served a fork of a lesson in order to get a modified version to the tablet I use while teaching. However, I don't know how common that is.

maxim-belkin:
> This would mean, though, that if someone wants to serve a fork of a lesson they would have to modify this file?  

this action only tests that websites can be built. GitHub would still build a website for your fork.

fmichonneau:
I agree, thanks Maxim

PR #507, opened 2020-10-08
update link to discuss mailing list

tobyhodges:
the current link target returns a 404.

I also suspect the email address and some other linkrefs need updating here...


fmichonneau:
Thanks Toby!

PR #508, opened 2020-10-16
update expected reference filename

tobyhodges:
Since `/bin/boilerplate/reference.md` has no `permalink` field, the generated page will be at `reference.html` not `reference/index.html` as expected by `make lesson-files`.


maxim-belkin (approved):
LGTM! Thanks, Toby!

PR #509, opened 2020-10-16
add default repository to install_required_packages()

zkamvar:
A lot of people do not have the default repository set in R, so this command fails in Make. Adding a default option fixes the issue.


maxim-belkin (approved):
Both explanation and implementation seem reasonable to me

fmichonneau:
thanks Zhian!

PR #510, opened 2020-10-21
lesson.scss: no borders around unrecognized code

maxim-belkin:
An example showing the border:
image

After the change:
image

Technically, selectors can be more general (`.highlight .err`) but more specific selectors shouldn't make things worse and might help avoid situations like 61db2b1.


fmichonneau:
Thanks!

PR #511, opened 2020-10-26
proposing a colour swap for hyperlinked text for more contrast

serahkiburu:
Because links are not underlined, I would like to propose a change in hyperlink colours so we can have more contrast with surrounding text. WebAIM suggests, that "For usability and accessibility, links should be underlined by default. Otherwise, link text must have at least 3:1 contrast with surrounding body text, and must present a non-color indicator (typically underline) on mouse hover and keyboard focus."

Here's a helpful tool to check for contrast ratios: https://webaim.org/resources/linkcontrastchecker/


zkamvar:
Should we then have links underlined?

We can specify it with the `text-decoration: underline` property

maxim-belkin:
> and must present a non-color indicator (typically underline) on mouse hover and keyboard focus."

> Should we then have links underlined?

Our links are underlined upon mouse hover

zkamvar:
> Our links are underlined upon mouse hover

I think Serah's enhancement would help us bring the template up to accessibility guidelines https://webaim.org/techniques/hypertext/link_text#appearance for links that are not underlined by default.

fmichonneau:
yes, it seems that using underline would improve accessibility

fmichonneau:
superseded with #549

Issue #513, opened 2020-11-04
Option to add a border around images?


tobyhodges:
In the process of developing a lesson using the remote theme, we ran into some trouble with screenshots embedded in the body of an episode: we had a short paragraph sandwiched between two screenshots, both of which showed blank/dark grey text on a white background (they are screenshots of the GitHub interface). One developer noticed that it was very easy to miss that paragraph between the two screenshots and that got us to wondering how we might more clearly mark out figures from text in the lesson.

Would it be possible to add a mechanism to place a border/shadow around images displayed on lesson pages? This might not be desirable for every lesson, or indeed even every image within a given lesson, but would be very helpful in cases like ours. Given the topics covered in Carpentries lessons, I'm sure we're not the only people using screenshots in our lesson, so I suspect this would help others too.

fmichonneau:
maybe we could create a new CSS class that would take care of this?

maxim-belkin:
The solution is to add CSS specification right after the image in question, e.g.:

```
![Alt text](image.svg){: style="border: solid 1px red; background: #eee;"}
```
It might be possible to add class specification as well (such as `{: .image-with-borders}`) and then define that class in `assets/css/lesson.scss` but I haven't tested this yet.

fmichonneau:
Toby implemented this in #514

PR #514, opened 2020-11-05
add class for images with a drop shadow

tobyhodges:
To address the issue described in #513, this introduces a `img.image-with-shadow` class with a drop shadow. If and when this is accepted, I'll document it in the lesson example.

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -302,7 +302,7 @@ span.fold-unfold {
 
 
 img.image-with-shadow {
-  box-shadow: 0 6px 24px 0 rgba(0, 0, 0, 0.25);
+  box-shadow: 0 6px 24px 0 #888;
 }
 
 //----------------------------------------
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -302,7 +302,7 @@ span.fold-unfold {
 
 
 img.image-with-shadow {
-  box-shadow: 0 4px 16px 0 rgba(0, 0, 0, 0.2), 0 6px 24px 0 rgba(0, 0, 0, 0.19);
+  box-shadow: 0 6px 24px 0 rgba(0, 0, 0, 0.25);
 }
 
 //----------------------------------------
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -301,6 +301,7 @@ span.fold-unfold {
 }
 
 
+p.image-with-shadow img,
 img.image-with-shadow {
   box-shadow: 0 6px 24px 0 #888;
 }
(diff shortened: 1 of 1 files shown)


assets/css/lesson.scss
```diff
@@ -299,6 +299,10 @@ span.fold-unfold {
 }
 
 
+img.image-with-shadow {
+  box-shadow: 0 4px 16px 0 rgba(0, 0, 0, 0.2), 0 6px 24px 0 rgba(0, 0, 0, 0.19);
```
maxim-belkin: This adds two shadows, both of which have no horizontal offset, similar vertical offsets (4 vs 6), different blur (16 vs 24), no spread, and very similar colors (~20% transparent black).

I'd suggest using a single shadow that would mimick the effect of these two shadows. It might also be easier on browsers to use some non-transparent gray color instead of transparent black (e.g. `#888`)

Another comment: we could apply this shadow by default to all images in `article` (by changing the selector from `img.image-with-shadow` to `article img`). What do you think?

maxim-belkin:
@tobyhodges, would you mind updating the PR to have a single drop shadow only? Or is there a case where two similar shadows are useful? Also, what do you think about changing `img.image-with-shadow` to `article img` ?

tobyhodges:
This fell off my radar, sorry. I have made a note to update the branch after the Core Team retreat is over (next week)

tobyhodges:
@maxim-belkin updated to use a single, grey shadow as you proposed. I did not adjust the selector to apply this shadow to all images in the `article` because I suspect the shadow effect will be undesirable in many circumstances. IMO it is better as an "opt-in" feature, which the `.image-with-shadow` class will provide.

maxim-belkin:
Thanks, Toby. Looks good. Current selector works when used like so:

```markdown
![alt text](../fig/figure.png){: .image-with-shadow}
![alt text](../fig/figure2.png){: class="image-with-shadow"}
```

I have to admit that in my first attempt I did:

```kramdown
![alt text](../fig/figure.png)
{: .image-with-shadow}
```

which resulted in the following HTML structure: `p[class="image-with-shadow"] > img`. I know this is my mistake but I wonder if this is something we want to take care of. All we'd have to do is extend the selector to images within paragraphs with "image-with-shadow" class:

```css
p.image-with-shadow img,
img.image-with-shadow { ...
```

tobyhodges:
That is a great suggestion, @maxim-belkin, because I agree that many lesson authors are likely to try what you tried (it is also what my muscle memory tries to make me do every time!)

tobyhodges:
Addressed in 4b20368

maxim-belkin (approved):
Looks great! 👍🏻

maxim-belkin:
Thank you, Toby!

PR #515, opened 2020-11-05
_config.yml: mention Carpentries Incubator

maxim-belkin:
To complement https://github.com/carpentries/lesson-example/pull/307


fmichonneau:
thanks!

PR #516, opened 2020-11-09
Matlab -> MATLAB

tobyhodges:
As @gcapes points out in https://github.com/swcarpentry/matlab-novice-inflammation/pull/230#issuecomment-723922937 `.language-matlab` code blocks should be prefaced with "MATLAB" rather than "Matlab" as they are currently.


fmichonneau:
Thanks Toby!

PR #520, opened 2020-11-20
lesson.scss: wildcard selectors for code blocks

maxim-belkin:
carpentries/lesson-example#316 explains how to use syntax highlighting for non-core languages. This PR changes the CSS selectors for code blocks to use wildcards so that lesson developers don't have to add `source` class to their code blocks (via  `{: .source}`) to add nice borders and title.


fmichonneau:
thanks for this!

PR #522, opened 2020-11-22
Don't trim trailing spaces in markdown

unode:
This caused formatting issues when using Theia through Gitpod

More info at:
https://github.com/carpentries-incubator/building-websites-with-jekyll-and-github-or-gitlab/pull/96
https://github.com/gitpod-io/gitpod/issues/896#issuecomment-722853235

Originally submitted at: https://github.com/carpentries/lesson-example/pull/310

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,7 @@ trim_trailing_whitespace = true
 indent_size = 2
 indent_style = space
 max_line_length = 100  # Please keep this in sync with bin/lesson_check.py!
+trim_trailing_whitespace = false  # keep trailing spaces in markdown - 2+ spaces are translated to a hard break (<br/>)
 
 [*.r]
 max_line_length = 80


maxim-belkin:
Hi @unode. Thanks for the PR!

I don't mind keeping whitespaces but I _don't_ think this is something we should be enforcing. In fact, GitHub lets you hide such changes when reviewing pull requests  -- click on the gear icon, select "Hide whitespace changes", and click "Apply and reload":

<details>
<summary>Screenshots</summary>

## The option
image

## Its effect
image

</details>

Also, I think trailing whitespaces have no effect on the produced webpages, so, if other maintainers of this repo decide to merge this, I'd suggest to update or remove the comment.

Note that I can't comment on how things work in GitPod -- it looks very interesting but currently The Carpentries doesn't use / recommend it for lesson development.

unode:
> Also, I think trailing whitespaces have no effect on the produced webpages, so, if other maintainers of this repo decide to merge this, I'd suggest to update or remove the comment.

I'm not sure if this is a lesser known feature but a double trailing space in markdown is significant and is translated into a [hard line break](https://github.github.com/gfm/#hard-line-breaks).

This can't be replicated in GitHub because the markdown flavour here renders soft line breaks as `<br/>`.

However, kramdown does follow the [soft/hard break distinction](https://kramdown.gettalong.org/syntax.html#paragraphs) and the rendered result is different (notice the highlighted 2 trailing spaces in the first screenshot):

screenshot_2020-12-10_03-41-58_670170497screenshot_2020-12-10_03-41-31_119080832

maxim-belkin:
Oh, interesting!  I wasn't aware of this double whitespace feature in kramdown. Thanks for sharing this.

I think we need to discuss this more. Is this feature present in the regular Markdown? If it is, then I'd vote to preserve whitespaces as you suggest. Otherwise, I wonder if trimming whitespaces would be a better choice because even though we use kramdown we might want to be closer to the plain-vanilla Markdown. What do you think? Do you rely on this feature in your lessons?

If we decide to preserve whitespaces, we might want to document/warn lesson developers about it in carpentries/lesson-example.

unode:
> Is this feature present in the regular Markdown?

[Yes](https://daringfireball.net/projects/markdown/syntax#p).

> Do you rely on this feature in your lessons?

I can't say I use this often in lessons but I do use it regularly when editing Markdown.
We also included this syntax [in one exercise](https://github.com/carpentries-incubator/jekyll-pages-novice/blob/gh-pages/_extras/more_markdown.md#solution) of the [WIP Jekyll lesson](https://github.com/carpentries-incubator/jekyll-pages-novice/).

maxim-belkin:
Oh, great! Let's update the comment then to clarify that we're talking about  2+ spaces, e.g.: 
> \# preserve trailing whitespace: 2+ trailing spaces are translated to a hard break

And again, I think we should document this in carpentries/lesson-example. Though, because you've got this info in your lesson, we can also point people to your lesson instead  "for more information on Markdown, see episode X of this lesson".

maxim-belkin (approved):
Thanks, @unode! 🥇

maxim-belkin:
Thanks, Renato. This change makes sense to me (and I learned something new) so I approve it. I do want to give other maintainers time (a day or two) to review the conversation and respond or merge the PR (if they approve of the changes too).

Again, thank you!

unode:
I've updated the comment in the commit.

As for the lesson, I've created [an issue on the Jekyll lesson](https://github.com/carpentries-incubator/jekyll-pages-novice/issues/167) to improve the exercise to better highlight this syntax and its pitfalls.

Waiting for others to review sounds good to me. Thanks a lot for the review and happy to share this knowledge :smiley_cat:

maxim-belkin:
Thanks for your contribution, @unode!

PR #527, opened 2020-12-02
R package installation: set CRAN url if default is "@CRAN@"

### LINKED ISSUE #526: No Default CRAN
When trying to use `make serve` to preview changes locally, I was receiving the following error:
``` 
Error in contrib.url(repos, "source") :
    trying to use CRAN without setting a mirror
Calls: install_required_packages -> install.packages -> contrib.url
Execution halted
make: *** [install-rmd-deps] Error 1
```

I was able to resolve by setting my default CRAN mirror following a suggestion from @zkamvar by adding `options(repos = c(CRAN = "https://cran.rstudio.org"))` to my `.Rprofile`, but others may be running into similar issues.

zkamvar:
We naïvely assumed that if the user did not set a default repository, that the default would be NULL. This is not the case. The default repository is set to `"@CRAN@"` and is [used by `contrib.url()` to abort if the session is not interactive](https://github.com/wch/r-source/blob/3a380278179be209a96ac509f67edd85e52e0612/src/library/utils/R/packages.R#L837).

If R packages need to be installed, this will check the default repository value and replace it if it's "@CRAN@"

This will fix #526

### DIFF
--- a/bin/dependencies.R
+++ b/bin/dependencies.R
@@ -9,6 +9,12 @@ install_required_packages <- function(lib = NULL, repos = getOption("repos", def
     c("rprojroot", "desc", "remotes", "renv"),
     rownames(installed.packages(lib.loc = lib))
   )
+  # The default installation of R will have "@CRAN@" as the default repository, which directs contrib.url() to either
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks!

PR #528, opened 2020-12-07
run R-based lessons in forks

zkamvar:
This is a modification for #501

This is a way to prevent this action from building in forks of `carpentries/styles`, but allowing it to be built in forks of other lessons (unless they happen to end with "styles").


fmichonneau:
Thanks Zhian

PR #530, opened 2020-12-17
Ignore .jekyll-metatada

klieret:
The `.jekyll-metadata` file appears for incremental builds, see [here](https://jekyllrb.com/docs/configuration/incremental-regeneration/), i.e. when running `bundle exec jekyll serve   --incremental`. It's not currently part of the "official" commands in the `Makefile`, but very useful when making a lot of consecutive changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .ipynb_checkpoints
 .sass-cache
 .jekyll-cache/
+.jekyll-metadata
 __pycache__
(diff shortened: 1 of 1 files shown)


fmichonneau:
Thanks @klieret. Note that this file is introduced with Jekyll 4. GitHub Pages is only [using](https://pages.github.com/versions/) Jekyll 3.9.

PR #531, opened 2020-12-17
Speed up builds of R-based lessons

zkamvar:
R-based lessons might take a while to build because packages need to be compiled from source. RStudio Package Manager has compiled versions of packages for ubuntu distros starting with 16.04: https://packagemanager.rstudio.com/client/#/repos/1/overview

I've added the necessary magic in the actions yaml to make it work.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -7,6 +7,9 @@ jobs:
   build-website:
     if: ${{ !endsWith(github.repository, '/styles') }}
     runs-on: ubuntu-latest
+    env:
+      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }} 
     defaults:
       run:
         shell: bash
@@ -45,6 +48,7 @@ jobs:
         uses: r-lib/actions/setup-r@master
         with:
           r-version: 'release'
+          http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions"
 
       - name: Install needed packages
         if: steps.check-rmd.outputs.count != 0
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -8,7 +8,7 @@ jobs:
     if: ${{ !endsWith(github.repository, '/styles') }}
     runs-on: ubuntu-latest
     env:
-      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest"
+      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }} 
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


.github/workflows/website.yml
```diff
@@ -45,6 +48,7 @@ jobs:
         uses: r-lib/actions/setup-r@master
         with:
           r-version: 'release'
+          http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions"
```
maxim-belkin: I've noticed you use `ubuntu-16.04` here. The `ubuntu-latest` runner that we use in this action is actually Ubuntu 18.04. Does that affect R packages?

maxim-belkin (commented):
Using pre-compiled packages will definitely speed up installation of R-packages, so 👍🏻 from me.
However, it would be ideal to also use `actions/cache` action to actually cache packages once they're installed. If you can figure this out in this PR -- that'd be great. Otherwise, I think there is only one issue that needs to be addressed.

maxim-belkin:
And one more request -- it'd be great if you could use your fork for future PRs. This is the procedure that all of our contributors are using.

.github/workflows/website.yml
zkamvar: Good catch! that was me foolishly copy/pasting the UA string from another example.

zkamvar:
> However, it would be ideal to also use `actions/cache` action to actually cache packages once they're installed. If you can figure this out in this PR -- that'd be great. Otherwise, I think there is only one issue that needs to be addressed.

Good point! I don't think it should be that complex. I'll run some tests and add to the PR.

.github/workflows/website.yml
zkamvar: Actually, looking at the latest version of the standard check, it seems they don't need a modified UA string any more. https://github.com/r-lib/actions/blob/master/examples/check-standard.yaml

zkamvar:
> > However, it would be ideal to also use `actions/cache` action to actually cache packages once they're installed. If you can figure this out in this PR -- that'd be great. Otherwise, I think there is only one issue that needs to be addressed.
> 
> Good point! I don't think it should be that complex. I'll run some tests and add to the PR.

On second inspection, it already uses the cache mechanism

https://github.com/carpentries/styles/blob/ced4bf5f045e1ae7b28ec3ce1f6cf0cbe27a8437/.github/workflows/website.yml#L68-L74

maxim-belkin:
If you observe "slowness", try moving caching step before the installation step. My understanding is that caching step has to precede the installation step in order for it to have an effect.

fmichonneau:
Thanks Zhian and Maxim. I think the confusion about caching here comes from the fact we are doing 2 types of package installation:
- the needed dependencies to get the scripts that are common to the building process
- the needed dependencies for the lessons specifically.

Caching is enabled for the second part but not the first.

Regarding the Ubuntu versioning, given that GitHub does a [progressive roll out](https://github.com/actions/virtual-environments/issues/1816 ) when they upgrade which version is being referred to as "latest" and given the need to know which version is being used to rely on the RSPM, I think it'd be safer for us to specify the ubuntu version manually instead of relying on the latest tag.

I'm going to merge this and address these issues in a separate PR.

PR #535, opened 2020-12-18
deploy from "website" action

fmichonneau:
this is an initial attempt to include deployment as part of the GitHub Actions so R-based lessons can use it without having to create a separate one.

One issue is that in my tests, the conditions on line 96 were too strict and the step was always skipped, I haven't figure out why.

This bumps to the limits of my git knowledge but I approached this by creating an orphan branch, building the lesson there, and force pushing this orphan branch to `gh-pages`. Suggestions for a better approach are welcome.

This also assumes that R-based lesson use `main` as their default branch (which is the case for most at the time but not all)

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RSPM: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"
-      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }} 
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     defaults:
       run:
         shell: bash
@@ -82,6 +82,24 @@ jobs:
           done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "18.04"), sep = "\n")')
 
       - run: make site
+
+      - name: Setup Git
+        if: ${{ steps.check-rmd.outputs.count != 0 }}
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          git status
+
+      - name: Push site
+        if: ${{ github.event.push == 'true' && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/head/main' }}
+        run: |
+          git checkout --orphan gh-pages
+          git add .
+          git commit -m "[GitHub Actions] render website"
+          git remote -v
+          git status
+          git push origin gh-pages
+
       - run: make lesson-check
         if: always()
       - run: make lesson-check-all
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -1,7 +1,9 @@
 name: Website
 on:
   push:
-    branches: gh-pages
+    branches:
+      - gh-pages
+      - main
   pull_request: []
 jobs:
   build-website:
@@ -88,17 +90,15 @@ jobs:
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "GitHub Actions"
-          git status
+          git checkout --orphan gha-rendering
 
       - name: Push site
-        if: ${{ github.event.push == 'true' && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/head/main' }}
+        if: ${{ github.event.push == 'true' && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/heads/main' }}
         run: |
-          git checkout --orphan gh-pages
-          git add .
+          rm -rf .bundle
+          git add --force .
           git commit -m "[GitHub Actions] render website"
-          git remote -v
-          git status
-          git push origin gh-pages
+          git push --force origin +gha-rendering:gh-pages
 
       - run: make lesson-check
         if: always()
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -83,24 +83,31 @@ jobs:
             eval sudo $cmd
           done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "18.04"), sep = "\n")')
 
-      - run: make site
+      - name: Render the markdown and confirm that the site can be built
+        run: make site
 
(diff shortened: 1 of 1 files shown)


zkamvar (commented):
Thank you for tackling this!

From my understanding of [the previous workflow](https://github.com/datacarpentry/r-raster-vector-geospatial/blob/master/.github/workflows/deploy.yml#L53-L57), you built the markdown files, and then moved everything except for the `_site/` folder into a separate folder called `_rendered` and used Max Held's action to push it to gh-pages.


I think an alternative might be to use the same strategy as {pkgdown} and create a worktree, effectively checking out the gh-pages branch in a different folder and moving the changes there. 

It would look something like this:

```sh
git worktree add --track -B gh-pages gh-pages origin/gh-pages
cp -r `ls -A | grep -v '.git' | grep -v '_site'` gh-pages
cd gh-pages
git add -A
git commit --allow-empty -m "[Github Actions] render website"
git push origin gh-pages
cd ../
git worktree remove --force gh-pages
```

zkamvar:
I tried this method out on [my fork of r-socialsci](https://github.com/zkamvar/r-socialsci) and can confirm that [the website](https://zkamvar.github.io/r-socialsci/04-ggplot2/index.html) does actually build successfully: https://github.com/zkamvar/r-socialsci/tree/gh-pages

Again, here are the steps:

1. setup git and create a folder called `gh-pages` that contains the `gh-pages branch`
2. build the site
3. copy everything except the `gh-pages`, `.git`, and `_site` folders to the `gh-pages` folder.
4. move into the `gh-pages` folder, add, commit, and push
5. move back into the main repo and remove the git worktree

```yaml
      - name: Set up Git
        if: ${{ github.event_name == 'push' && steps.check-rmd.outputs.count != 0 }}
        run: |
          git config --local user.email "actions@github.com"
          git config --local user.name "GitHub Actions"
          git worktree add --track -B gh-pages gh-pages origin/gh-pages
          ls -larth | tail
      - run: make site
      - name: Commit
        if: ${{ github.event_name == 'push' && steps.check-rmd.outputs.count != 0 }}
        run: |
          cp -r `ls -A | grep -v 'gh-pages' | grep -v '.git' | grep -v '_site'` gh-pages
          cd gh-pages
          git add -A .
          git commit --allow-empty -m "[Github Actions] render website"
          git push origin gh-pages
          cd ../
      - name: Tear Down
        if: ${{ github.event_name == 'push' && steps.check-rmd.outputs.count != 0 }}
        run: |
          git worktree remove --force gh-pages

```

.github/workflows/website.yml
```diff
@@ -82,6 +84,22 @@ jobs:
...
+
+      - name: Setup Git
+        if: ${{ steps.check-rmd.outputs.count != 0 }}
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          git checkout --orphan gha-rendering
+
+      - name: Push site
+        if: ${{ github.event.push == 'true' && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/heads/main' }}
```
zkamvar:
```suggestion
        if: ${{ github.event_name == "push" && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/heads/main' }}
```

.github/workflows/website.yml
```diff
@@ -1,15 +1,17 @@
 name: Website
 on:
   push:
-    branches: gh-pages
+    branches:
+      - gh-pages
+      - main
```
zkamvar, line 6: we can also add the "master" branch here and condition the build yaml on the ref being main or master.

.github/workflows/website.yml
```diff
@@ -82,6 +84,22 @@ jobs:
...
+        if: ${{ github.event.push == 'true' && steps.check-rmd.outputs.count != 0 && github.ref == 'refs/heads/main' }}
+        run: |
+          rm -rf .bundle/ _site/
+          git add --force .
+          git commit -m "[GitHub Actions] render website"
+          git push --force origin +gha-rendering:gh-pages
+
       - run: make lesson-check
         if: always()
       - run: make lesson-check-all
```
zkamvar: Ignore my previous suggestions to use worktrees. 

This solution does work, but instead of checking out a dummy orphan branch and forcing it to be the new github pages branch, we checkout github pages into a separate folder with the checkout action and copy the rendered files there. 

This proposal also keeps the history of the changes since these can change due to the R environment.

```suggestion
      - name: Render the markdown and confirm that the site can be built
        run: make site

      - name: Checkout github pages
        if: ${{ github.event_name == 'push' && steps.check-rmd.outputs.count != 0 && github.ref != 'refs/heads/gh-pages'}}
        uses: actions/checkout@master
        with:
          ref: gh-pages
          path: gh-pages

      - name: Commit and Push
        if: ${{ github.event_name == 'push' && steps.check-rmd.outputs.count != 0 && github.ref != 'refs/heads/gh-pages'}}
        run: |
          # copy everything into gh-pages site
          cp -r `ls -A | grep -v 'gh-pages' | grep -v '.git' | grep -v '.bundle/' | grep -v '_site'` gh-pages
          # move into gh-pages, add, commit, and push
          cd gh-pages
          # setup git
          git config --local user.email "actions@github.com"
          git config --local user.name "GitHub Actions"
          git add -A .
          git commit --allow-empty -m "[Github Actions] render website (via ${{ github.sha }}"
          git push origin gh-pages
          # return
          cd ..

      - run: make lesson-check-all
```

zkamvar (changes requested):
I have suggested changes that would allow for history to be kept AND the source commit to be recorded. I think this makes things a bit clearer instead of using worktrees or orphan branches with force pushes.

I've also added a suggestion to only build when the pushed branch is not gh-pages (though it seems like all of the R repositories that render their results are already on alternative branches).

Also, to keep a trail, I've added the commit sha to the gh-pages commit. You can see an example of how it turns out at https://github.com/zkamvar/r-socialsci/commit/fc3fd1d74c74cac7ca91b436f988e9c65911d357

I've also removed make check since that was redundant with make-lesson-check-all.

PR #537, opened 2020-12-22
bin/lesson_check.py: allow 'caution' blockquote

maxim-belkin:
Fix for `Unknown or missing blockquote type caution`

CC @zkamvar


zkamvar (approved):
LGTM and looks like you got to this before I did! Good work :)

PR #540, opened 2021-01-07
pin ubuntu version to 20.04

fmichonneau:
given that GitHub Actions transitions progressively from 18.04 to 20.04 and that we need to know the version nickname to set the RSPM repo correctly, I'd like to propose we pin the ubuntu version to 20.04 and upgrade manually when needed.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -8,9 +8,9 @@ on:
 jobs:
   build-website:
     if: ${{ !endsWith(github.repository, '/styles') }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     env:
-      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"
+      RSPM: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     defaults:
       run:
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -81,7 +81,7 @@ jobs:
           while read -r cmd
           do
             eval sudo $cmd
-          done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "18.04"), sep = "\n")')
+          done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "20.04"), sep = "\n")')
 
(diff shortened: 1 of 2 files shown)


maxim-belkin (approved):
👍🏻  Seems reasonable given that we use Ubuntu's codename for RSPM

zkamvar:
There's a weird error with stringi not being found for the R intro Geospatial, but I'm not sure why it's happening as stringi is a dependency of one of the other packages and should be installed: 

```
Error in dyn.load(file, DLLpath = DLLpath, ...) : 
  unable to load shared object '/home/runner/work/_temp/Library/stringi/libs/stringi.so':
  libicui18n.so.60: cannot open shared object file: No such file or directory
Calls: source ... namespaceImport -> loadNamespace -> library.dynam -> dyn.load
```

zkamvar:
I'm going to rerun the tests to see if it's stochastic, but otherwise, I also approve. It makes sense for us to do this.

fmichonneau:
@zkamvar I think we need to add the os version number into the [string that gets hashed](https://github.com/carpentries/styles/pull/540/files#diff-c7731b29de71fca225e1dfac88ebe31975a4cd9759ab13e91d86eeadafbddc10L75)

zkamvar:
We don't have a matrix run, so d4b947f would return an unspecific empty string.

.github/workflows/website.yml
```diff
@@ -72,16 +72,16 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ env.R_LIBS_USER }}
-          key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
-          restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
+          key: ${{ matrix.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
+          restore-keys: ${{ matrix.os }}-${{ hashFiles('.github/R-version') }}-1-
```
zkamvar:
```suggestion
          key: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-${{ hashFiles('.github/depends.Rds') }}
          restore-keys: ${{ runner.os }}-${{ hashFiles('.github/R-version') }}-1-
```

fmichonneau:
hacky but works for now.

Issue #544, opened 2021-02-08
[GitHub Actions] 'Website' workflow keeps deleted files in `gh-pages` branch


fmichonneau:
As `gh-pages` isn't an orphaned branch anymore, files that are deleted in `main` remain in the rendered version in `gh-pages`. The workflow needs to be adapted to make sure these files are also removed.

zkamvar:
This is a good point, and I believe it is relevant with the {sandpaper} workflows as well.

PR #545, opened 2021-02-08
add patch to clean gh-pages before committing

zkamvar:
This will address #544 by cleaning up all previous files before adding new ones to gh-pages


zkamvar:
This should work, assuming that there are not auto-generated things in the main branch.

fmichonneau (approved):

PR #548, opened 2021-02-19
Change ruby executable to "bundle exec ruby".

### LINKED ISSUE #547: Python invokes native ruby, not bundle ruby
https://github.com/carpentries/styles/blob/ec29c6b1793ed0af52b4c858a0c7a4911722caf1/bin/util.py#L118

The Ruby invoked by the Python tools called from the Makefile runs a different Ruby than when Ruby is called directly from the Makefile.

For example, the "serve" target runs a multi-stage workflow that populates the bundle with the required gems, and then runs jekyll.

The Python code just calls `ruby` via Popen.

These executables therefore run in different environments, and can have inconsistent behavior, which is hard to debug.

Recommend changing the `ruby` invocation on the line above to `bundle exec ruby`.

reid-a:
Closes #547

### DIFF
--- a/bin/util.py
+++ b/bin/util.py
@@ -115,7 +115,7 @@ def read_markdown(parser, path):
              for (i, line) in enumerate(body.split('\n'))]
 
     # Parse Markdown.
-    cmd = 'ruby {0}'.format(parser)
+    cmd = 'bundle exec ruby {0}'.format(parser)
     p = Popen(cmd, shell=True, stdin=PIPE, stdout=PIPE,
               close_fds=True, universal_newlines=True, encoding='utf-8')
     stdout_data, stderr_data = p.communicate(body)


maxim-belkin:
Hi @reid-a! Thanks for the PR. Could you please provide a reproducer where `ruby --version` and `bundle exec ruby --version` report different results?

reid-a:
Hi @maxim-belkin -- 

I am happy to engage!

This issue arose in the context of the hpc-carpentry lesson development, where changes to the markdown parsing were present in the bundle but not in the outermost Ruby environment. There's a lot of infrastructure that's implicated in the problem, can you clarify what the starting point would be for the reproducer?

The most basic indication is, at the command-line in the repo, you can see differences in the gem libraries after the kramdown-parser-gfm gem was added as a requirement to the Gemfile, viz:
```
$ gem list | grep kramdown
kramdown (2.3.0, 1.17.0)
$ bundle list | grep kramdown
  * kramdown (2.3.0)
  * kramdown-parser-gfm (1.1.0)
```
Since the gfm parser is now a requirement for us, the result of this is that, prior to this PR change, `make lesson-check-all` will fail with a load error trying to load the gfm parser, but the CI will pass -- this inconsistent behavior makes debugging the lesson hard, and motivates the fix.

This is motivation, more than reproduction, though.

reid-a:
Just to clarify, the thing actually asked for is not true, the ruby executables are the same version between the two cases.

maxim-belkin:
> Just to clarify, the thing actually asked for is not true, the ruby executables are the same version between the two cases.

👍🏻  I'm writing a more detailed response (it'll take me some time) but wanted to quickly respond to this note: yes, Ruby executable should be the same and that's what confused me in your report in #547.

maxim-belkin:
> can you clarify what the starting point would be for the reproducer?
> 
> The most basic indication is, at the command-line in the repo, you can see differences in the gem libraries after the kramdown-parser-gfm gem was added as a requirement to the Gemfile, viz:
> 
> ```
> $ gem list | grep kramdown
> kramdown (2.3.0, 1.17.0)
> $ bundle list | grep kramdown
>   * kramdown (2.3.0)
>   * kramdown-parser-gfm (1.1.0)
> ```

The code block above is a good starting point and the output you get makes sense: `github-pages` gem that we require in the Gemfile includes `kramdown-parser-gfm` and our Makefile instructs Bundler to install all gems into `.vendor` directory. The `gem` command knows nothing about this `.vendor` directory, so it doesn't see gems installed there. So, it's not the Ruby version that is different, as you've noticed already, but the fact that `gem` command knows nothing about the `.vendor` directory. 

The next question is whether the proposed change is sufficient. There are two issues with `bundle exec ruby`: it relies on Bundler being installed and assumes that `bundle install` (with proper flags) has already been executed. If either of these is not the case, `make lesson-check` will error out with a long error message.

Anyway, I think this is a good quick fix and we should merge it. Meanwhile, I'm going to prepare another fix that should take care of the two edge cases I mentioned above.

Thank you for your contribution to carpentries/styles, Andrew!

maxim-belkin (approved):

reid-a:
Thanks @maxim-belkin !

You're right that this is a short-term fix.

As for the sufficiency issues, you're right (IMO) about the two issues (bundler must be present, and the new "lesson-check-all" relies on order-of-operations).

The first issue is not local to the change, the Makefile as a whole assumes this, so this needs to be addressed in a broader scope.

The order-of-operations issue is potentially fixable locally by changing the "bundle exec ruby" to do the whole build chain, emulating the existing Jekyll targets (`make serve`, `make site`). I didn't go that route, obviously, but one could

Issue #552, opened 2021-03-03
Using ``make serve`` with ruby 3.0.0


josephmje:
I had an issue running `make serve` using ruby 3.0.0. I tracked down a solution on this GitHub issue: https://github.com/jekyll/jekyll/pull/8590.

The fix was to run `bundle add webrick`.

This command adds the following line to the Gemfile:
`gem "webrick", "~> 1.7"`

Should the [Carpentries lesson setup instructions](https://carpentries.github.io/lesson-example/setup.html) be modified to explicitly install ruby 2.7.1 or would a change to the Gemfile be warranted to add the webrick dependency?

maxim-belkin:
Thanks for the report, @josephmje! Our Gemfile currently imposes `>= 2.7.1` requirement on Ruby. Below are the possible solution that I envision:

1. Require Ruby version == 2.7.1
2. Require Ruby version >= 2.7.1 < 3.0.0
3. Add `webrick` requirement for Ruby version 3.0.0 and above:
```ruby
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.0.0')
    gem "webrick", "~> 1.7"
end
```
I think the latter option is probably the cleanest one (because it does not add new dependencies for Ruby versions prior to 3.0.0 while still allowing people to use Ruby 3.0.0)

josephmje:
Yes, I agree. I think there are a lot of people that use Ruby 3.0.0. How about

`gem 'webrick', '~>1.6'`

since 1.6 is the version bundled in Ruby 2.7.1 and it would also allow anything below 2.0?

maxim-belkin:
From Ruby 3.0 release notes (https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/):

> The following libraries are no longer bundled gems or standard libraries. Install the corresponding gems to use these features.
> * sdbm
> * webrick
> * net-telnet
> * xmlrpc

It looks like there is a vulnerability in 1.6 ([link](https://www.ruby-lang.org/en/news/2020/09/29/http-request-smuggling-cve-2020-25613/)), so we should probably stick with `>= 1.6.1`.
Webrick is a default gem for Ruby 2.7.1 and 2.7.2, so there is no need to require it for versions prior to 3.0.0... but, of course, there should be no harm if we require it unconditionally too.

josephmje:
Good catch @maxim-belkin! This sounds good to me.

maxim-belkin:
[...]

PR #555, opened 2021-03-05
Adds webrick dependency if ruby version less than 3.0.0

josephmje:
Discussed in #552. `webrick` is no longer bundled with Ruby as of version 3.0.0 so this PR adds a check to determine which Ruby version is being used and install `webrick` if necessary.

Would you suggest also testing against Ruby 3.0 in the Github Actions?


maxim-belkin (approved):
Thank you, Michael!

maxim-belkin:
Thank you for the report and contribution, Michael! 👍🏻

josephmje:
Thank you @maxim-belkin for helping with the possible solutions!

PR #557, opened 2021-03-10
lesson_check.py allow for missing life_cycle

### LINKED ISSUE #556: Error in lesson_check.py if lesson does not have life_cycle
Source: https://github.com/swcarpentry/r-novice-gapminder/pull/694#issuecomment-795848265

If a lesson does not have `life_cycle` (e.g. all of the official lessons), then `lesson_check.py` fails. This is a new bug introduced in https://github.com/carpentries/styles/pull/536. The solution is to check if the key "life_cycle" exists and return None if it doesn't.

zkamvar:
There was an issue with make lesson-check on the core lessons that didn't have the life_cycle parameter set. This PR fixes the issue by allowing `life_cycle` to be missing in the _config.yaml (which is standard for core lessons).

Notably, this was not caught in our earlier checks because our github workflow only tested if the sites could be built, but never checked their validity. 

This will fix https://github.com/carpentries/styles/issues/556

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -188,7 +188,10 @@ def check_config(reporter, source_dir):
         reporter.check(defaults in config.get('defaults', []),
                    'configuration',
                    '"root" not set to "." in configuration')
-    return config['life_cycle']
+    if 'life_cycle' in config:
+        return config['life_cycle']
+    else
+        return None
 
 def check_source_rmd(reporter, source_dir, parser):
     """Check that Rmd episode files include `source: Rmd`"""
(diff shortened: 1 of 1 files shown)


maxim-belkin:
The fix looks good to me. It might be slightly better to set `config['life_cycle']` and then return it just in case we want to catch this edge case somewhere else.

```python
    if 'life_cycle' not in config:
        config['life_cycle'] = None

    return config['life_cycle']
```

zkamvar:
Thank you! I made the change and will merge once the checks pass.

PR #561, opened 2021-03-18
Makefile: silence Docker commands

maxim-belkin:
Put `@` in front of Docker commands to silence them up (we do so everywhere else).


zkamvar (approved):
LGTM

PR #562, opened 2021-03-18
Makefile: use SHELL to call bin/knit_lesson.sh

maxim-belkin:
Use `SHELL` to call `bin/knit_lessons.sh` script. This should let us avoid relying on the executable bit of the script.


zkamvar (approved):
This looks okay to me. I know that our scripts assume BASH, but modern macOS uses zsh, so I tried checking the scripts with `shellcheck --shell sh`and the only one that doesn't pass the test is `bin/run-make-docker-serve.sh`, which we no longer use (and should be removed)

PR #566, opened 2021-03-23
Makefile: don't fail when Python isn't found

maxim-belkin:
Python shouldn't be a hard requirement in our Makefile -- there are targets that don't need Python at all (serve, site, clean, docker-serve, etc.) Instead, make it a hard requirement for targets that need it.


zkamvar (approved):
YES! This is a major pain point for a lot of folks and removing this barrier would make the process a lot easier

maxim-belkin:
Thanks, Zhian!

PR #567, opened 2021-03-27
bin/dependencies.R: handle 'no packages were specified' error

maxim-belkin:
Fixes the following issue:

```
$ make site
lib paths: /Library/Frameworks/R.framework/Versions/3.5/Resources/library
Error in install.packages(missing_pkgs, lib = lib, repos = repos) :
  no packages were specified
Calls: install_required_packages -> install.packages
Execution halted
make: *** [install-rmd-deps] Error 1
```


zkamvar (approved):
Thank you for catching that!

Issue #154, opened 2017-06-30
provide mathjax


fmichonneau:
There a few lessons (in the Reproducible curriculum lessons) that use math notation and require MathJax, but it is currently not part of the template.

rgaiacs:
This is related with https://github.com/swcarpentry/DEPRECATED-bc/issues/521 and https://github.com/swcarpentry/DEPRECATED-lesson-template/issues/121.

@fmichonneau Could you provide me with the link of the episodes that you mentioned? As I mentioned before, I believe that we still can avoid MathJax.

twitwi:
I've been using equation-in-a-browser for some time (for my slide decks) and I switched to [katex](https://khan.github.io/KaTeX/) as mathjax was very very annoying to bundle and heavy to host or for offline use. Katex is pretty active and has good math support now.

To give an idea, the Katex download is less than 2MB. By keeping only [fonts for decents browsers](http://caniuse.com/#search=woff), we can probably go down to around 300kB (unzipped).

rgaiacs:
@twitwi @fmichonneau I have think about it in the past last days. What do you think of require that pages that are going to use math, only for a math variable or for a very complex equation, **must** include

~~~
require_mathjax: true
~~~

in the YAML header?

This will allow us to trigger the inclusion of MathJax or KaTeX only on pages that need it.

fmichonneau:
I think there are 2 separate issues here:

- supporting a way to render math notations in our lessons
- providing support for this method offline

For this first, I think it would be good to have some kind of support, as currently there is none. I don't know much about the topic, so I'll let people with more knowledge decide whether MathJax, KaTeX, or something else is the best option. 

For the second issue, I don't think we should worry about providing offline support for this. It is nice that the current template doesn't need an internet connection to look good, but there are only a handful of places where equations are used/needed, and it doesn't seem that having offline support for those is worth adding a hard coded dependency to our template.

twitwi:
@rgaiacs  Something like `require_mathjax` is a good idea to limit the footprint of most pages (that don't use any latex).
I'd go for something more generic though, like `contains_latex_math` so we can swap implementations without breaking the semantics of the flag.

I agree that it is most important to have it work online. Offline matters only secondarily.
To know if what we can use, @fmichonneau do you have the link to a lesson that uses latex?

fmichonneau:
[...]

PR #573, opened 2021-04-14
add math support with katex

alanocallaghan:
https://github.com/carpentries/carpentries-theme/pull/13

#154


alanocallaghan:
https://github.com/carpentries/carpentries-theme/pull/13

alanocallaghan:
I'll be real I have no idea what I'm doing with katex nor with yarn/npm/etc so I'm not gonna try to do this

alanocallaghan:
Aight I think I made it work. I have no idea how to check that the implementation in this repo is correct (?) but this is the same code that works in a remote theme or when modifying one of the local files so should be equivalent

@fmichonneau

fmichonneau:
Thank you for spending time figuring it out. This looks good! I'll merge and add to the remote theme as well so we can test further.

PR #578, opened 2021-04-16
Improved relative_root_path

maxim-belkin:
Fix for relative_root_path that doesn't require Python intervention. Below I copy the explanation I added to `base_path.html`:

When the website is built by GitHub Pages, 'site.url is set to 'https://username.github.io', and 'site.baseurl' is set to '/lesson-name'. In this case we set 'relative_root_path' to 'site.url + site.baseurl'.

When we start a local server using `jekyll serve`, 'site.url' is set to 'http://localhost:4000' and 'site.baseurl' is empty. In this case we set 'relative_root_path' to 'site.url'.

When we build a website locally with `jekyll build`, both 'site.url' and 'site.baseurl' are empty. This case is handled by the last 'else' in the code below. The logic there follows the (adapted) instructions found at: https://ricostacruz.com/til/relative-paths-in-jekyll,     `page.url` gives the URL of the current page with a leading /:

  - when the URL ends with an extension (e.g., /foo/bar.html),
    we can get the 'depth' of the page by counting the number of
    forward slashes ('/') and subtracting 1
  - when the URL ends with a forward slash (e.g. /foo/bar/),
    we can get the depth of the page by counting the number of /

### DIFF
--- a/_includes/base_path.html
+++ b/_includes/base_path.html
@@ -1,27 +1,41 @@
-{% comment %}
-This is adapted from: https://ricostacruz.com/til/relative-paths-in-jekyll
+{%- comment -%}
+When the website is built by GitHub Pages,
+'site.url' is set to 'https://username.github.io'
+'site.baseurl' is set to '/lesson-name'
 
-`page.url` gives the URL of the current page with a leading /:
+When we start a local server using `jekyll serve`,
+'site.url' is set to 'http://localhost:4000' and
+'site.baseurl' is empty.
 
-- when the URL ends with the extension (e.g., /foo/bar.html) then we can get
-  the depth by counting the number of / and remove - 1
-- when the URL ends with a / (e.g. /foo/bar/) then the number / gives the depth
-  directly
-{% endcomment %}
+In both of the above cases we set 'relative_root_path' to 'site.url + site.baseurl'.
 
-{% assign relative_root_path = '' %}
+When we build a website locally with `jekyll build`,
+both 'site.url' and 'site.baseurl' are empty.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
actually, I think I'll combine the first two cases into one because `site.url + site.baseurl` == `site.url` when we do `jekyll serve`.

zkamvar:
I will take a look at this later today. It looks good so far. Thank you for wrestling this bear that is Liquid templating.

zkamvar:
I just took a look at this using a newly generated site and it works! 

screen capture of 404 page from the url http://localhost:4000/01-introduction/a/a/a/a/a/a/a/a/a/a


I especially like your solution for getting the appropriate page depth redirect.

maxim-belkin:
Thanks for merging, Zhian!

jhlegarreta:
Hooray, thanks @maxim-belkin, @zkamvar and @unode  for the effort !

What changes should be made in the repositories to have #568 fixed following these changes?
e.g. https://carpentries.github.io/lesson-example
or
https://github.com/carpentries-incubator/SDC-BIDS-dMRI

zkamvar:
@jhlegarreta 

The solution is to update the repo with the styles repository: https://github.com/carpentries/maintainer-resources/blob/main/FLIGHT_RULES.md#update-styles

I can go ahead and initiate both of these.

(also, to be clear: I was just a reviewer, @unode and @maxim-belkin definitely deserve the credit for working this out)

jhlegarreta:
Thanks for the explanations @zkamvar.

I see that you've already done that for me/us. Thanks :100:.

PR #584, opened 2021-04-21
Template workflow: add two more lessons

maxim-belkin:
I propose to add 2 more lessons to our template-testing matrix: carpentries/lesson-example and datacarpentry/astronomy-python. The first one is probably one of the lessons that the first-time maintainers look at pretty regularly. The latter one uses The Carpentries remote theme, which makes it a unique lesson in the build matrix.

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -8,14 +8,17 @@ jobs:
     name: ${{ matrix.lesson-name }} (${{ matrix.os-name }})
     if: github.repository == 'carpentries/styles'
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         lesson: [swcarpentry/shell-novice, datacarpentry/r-intro-geospatial, librarycarpentry/lc-git]
         os: [ubuntu-20.04, macos-latest, windows-latest]
+        experimental: [false]
+        remote-theme: [false]
         include:
           - os: ubuntu-20.04
-            os-name: Ubuntu
+            os-name: Linux
           - os: macos-latest
             os-name: macOS
           - os: windows-latest
@@ -26,6 +29,17 @@ jobs:
             lesson-name: (DC) R Intro Geospatial
           - lesson: librarycarpentry/lc-git
             lesson-name: (LC) Intro to Git
+          - lesson: datacarpentry/astronomy-python
+            lesson-name: (DC) Foundations of Astronomical Data Science
+            experimental: true
+            os: ubuntu-20.04
+            os-name: Linux
+            remote-theme: true
+          - lesson: carpentries/lesson-example
+            lesson-name: (CP) Lesson Example
+            experimental: false
+            os: ubuntu-20.04
+            os-name: Linux
     defaults:
       run:
         shell: bash # forces 'Git for Windows' on Windows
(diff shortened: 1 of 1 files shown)


ErinBecker:
Two thumbs way way up to including astronomy-python in the testing matrix. @rudyphd and I have been trying to get that lesson to stop failing checks for the past two weeks and failing . . . .

maxim-belkin:
The "Test template" action doesn't run `make lesson-check` or `make lesson-check-all`, so it wouldn't tell us if there are any issues with these checks for the tested lessons. Technically, we can add `make lesson-check-all` to the "Test template" action because it doesn't affect the final status of the action. But it'll come at a cost of about a minute of GitHub Action time per pull request. I'm not sure how much GH Action time The Carpentries have but I assume  2000 minutes -- perhaps someone could confirm this here: https://github.com/organizations/carpentries/settings/billing (I don't have access to this page).

zkamvar:
> The "Test template" action doesn't run `make lesson-check` or `make lesson-check-all`, so it wouldn't tell us if there are any issues with these checks for the tested lessons. Technically, we can add `make lesson-check-all` to the "Test template" action because it doesn't affect the final status of the action. But it'll come at a cost of about a minute of GitHub Action time per pull request. I'm not sure how much GH Action time The Carpentries have but I assume 2000 minutes -- perhaps someone could confirm this here: https://github.com/organizations/carpentries/settings/billing (I don't have access to this page).

It's 3000 minutes, but that's for private repos. It's currently unlimited for public repos. 

Given that, I think `make lesson-check-all` would definitely be good to add.

zkamvar:
That being said, would you like me to add the change to your branch?

maxim-belkin:
Sure

maxim-belkin:
I think we should add it as a separate step... what do you think?

zkamvar:
> I think we should add it as a separate step... what do you think?

Yes, it makes much more sense to do that so that we avoid mixing up the outputs of the two commands

.github/workflows/template.yml
```diff
@@ -134,5 +149,7 @@ jobs:
             eval sudo $cmd
           done < <(Rscript -e 'cat(remotes::system_requirements("ubuntu", "20.04"), sep = "\n")')
 
-      - run: make site
+      - run: |
+          make site
+          make lesson-check-all
```
zkamvar:
```suggestion
      - run: make site
      - run: make lesson-check-all
```

zkamvar:
Right now, the only one that's failing is Astronomy because of the switch to the remote theme (which needs to update the build tools separately).

maxim-belkin:
> the only one that's failing is Astronomy

This should've been fixed in carpentries/styles#585. I'll try closing and reopening this PR because force-pushes don't seem to help.

maxim-belkin:
OK, so once https://github.com/datacarpentry/astronomy-python/pull/28 is merged, the (DC) Foundations of Astronomical Data Science (Linux) check should pass

maxim-belkin:
> All checks have passed
1 skipped and 11 successful checks

🎉

zkamvar:
Thank you so much for this, @maxim-belkin! Yay for more robust integration tests and yay for finally getting GHA to work!

I will merge this now!

fmichonneau:
Thank you everyone for this!

PR #585, opened 2021-04-22
bin/lesson_check.py: one more fix for using_remote_theme()

maxim-belkin:
This PR "fixes" `make lesson-check` that currently fails for datacarpentry/astronomy-python.


zkamvar:
LGTM. Now we are both even for missing this argument :grin:

PR #588, opened 2021-04-23
Two more

maxim-belkin:
resubmitting #584 in an attempt to make GitHub happy.


maxim-belkin:
> resubmitting #584 in an attempt to make GitHub happy.

and GitHub is like: nope!

zkamvar:
This is a bug on the Astronomy lesson side. The Makefile needs to be updated there.

zkamvar:
I see you already did so. Please ignore my comment.

PR #589, opened 2021-04-23
Template workflow: Install R packages after restoring cache

maxim-belkin:
To have an effect, the caching step has to preceed the installation step.
Here, I move the step that installs some R packages to after the caching step.
I used [this](https://github.com/r-lib/actions/tree/master/examples#when-should-you-use-it-1) as an example.

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -109,13 +109,6 @@ jobs:
         with:
           r-version: 'release'
 
-      - name: Install needed packages
-        if: steps.check-rmd.outputs.count != 0
-        run: |
-          packages = setdiff(c('remotes', 'rprojroot', 'renv', 'desc', 'rmarkdown', 'knitr'), rownames(installed.packages()))
-          install.packages(packages, repo="https://cran.rstudio.com/")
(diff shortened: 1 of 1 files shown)


zkamvar:
There are three steps installing packages:

"Install needed R packages": our tools
"Query dependencies": individual lesson tools
"Install stringi from source": avoiding Ubuntu 18.04/20.04 screwups

They should be in that order and the restoring of cache should come before that.

maxim-belkin:
Ah, ok. I was looking at Linux build logs and 'stringi' is always built from source, so I thought the order is messed up.
I wonder if we can cache this step somehow -- the logs say

```
installing via 'install.libs.R' to /home/runner/work/_temp/Library/00LOCK-stringi/00new/stringi
```

maybe we could cache that path? ... anyways, closing the PR.

zkamvar:
You are correct that the cache step needs to come BEFORE all of the installation. The reason this PR is failing is because it came in the middle of the installation (I think).

In any case, the authors of the workflow acknowledged that they made a mistake when they initially named it "Cache R packages" instead of "Restore R package cache"

PR #592, opened 2021-05-03
Add katex fonts

alanocallaghan:
Trying to solve a problem of my own making. When using katex locally or on gh-pages, the math seems to render fine, but locally jekyll complains that
```
[2021-05-03 16:41:59] ERROR `/assets/css/fonts/KaTeX_Main-Regular.woff2' not found.
[2021-05-03 16:41:59] ERROR `/assets/css/fonts/KaTeX_Main-Regular.woff' not found.
[2021-05-03 16:41:59] ERROR `/assets/css/fonts/KaTeX_Main-Regular.ttf' not found.
```
which seems not great.

Getting Katex to use a reasonable font dir was "fun" ([...](https://github.com/KaTeX/KaTeX/issues/1987)) but I got there in the end.
 I think adding all of these is probably excessive but I'm not sure exactly which to add. Hence this being a draft until that's a bit more clear.


fmichonneau:
please ping when ready to merge! Thank you so much for working on this.

maxim-belkin:
Looks like we need all of these fonts. I checked out font-families at https://katex.org/docs/supported.html and was able to find quite a few of what is added in this PR.

One comment from me: I'm not sure if we're saving much by using minimized version of the JS, so maybe we could switch to the standard one?

maxim-belkin (approved):

alanocallaghan:
I think Calligraphic, Fraktur, SansSerif, Script, Typewriter are not necessary (as suggested here: https://github.com/KaTeX/KaTeX/issues/2088)

maxim-belkin:
I don't think that issue suggests that the fonts you mention are unnecessary. It says that if you know you don't need them... in our case we can't know in advance what fonts we need.

alanocallaghan:
I think you absolutely can know that by stating when describing katex support "we don't currently support X fonts but you can add them if you need them".

However if you want to add everything that's fine by me also

alanocallaghan:
Minified katex is 257kb rather than 594kb for the full version, should I switch them out? Every little helps long term if more libraries get added

PR #594, opened 2021-05-06
bin/lesson_check.py: allow exceptions to line length limit

### LINKED ISSUE #591: Allow skipping line length checks for link-only lines
The `make leson-check-all` script checks the line lengths of the Markdown files. If I'm not mistaken by the check results, lines containing only links dot not raise warnings if exceeding the allowed line length. However, such warnings are indeed raised for lines containing only links in the solution blocks.

Please correct me if I'm wrong, but I believe that links cannot be split across lines.

Below are some examples of such warnings:
The GitHub badge links on file
https://raw.githubusercontent.com/carpentries-incubator/SDC-BIDS-dMRI/main/README.md
pointed in
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2438761253?check_suite_focus=true#step:16:19

Line 456 on file
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/99dd1d1235e78bf940c4728974635fec5a00fabc/_episodes/constrained_spherical_deconvolution.md
pointed in
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2438761253?check_suite_focus=true#step:16:20

Line 314 on file
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/99dd1d1235e78bf940c4728974635fec5a00fabc/_episodes/deterministic_tractography.md
pointed in
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2438761253?check_suite_focus=true#step:16:21

Thus, ideally the checker should ideally skip such lines/avoid raising warnings for such lines even in the `solution` block quotes (or any other relevant block quotes).

Thanks.

maxim-belkin:
Allow lines that contain a single image or a single link to go over the suggested line length limit. So, all of the following lines will be allowed to go over the line limit:

```
[alt text](/some/url)
![alt text](/path/to/image)
[alt text][internal_link]
![alt text][internal_link_to_image]
```

Plus all of the above prefixed with any combination of `>` and ` ` to allow these lines to go over the limit in various code blocks.

Fixes carpentries/styles#591

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -56,6 +56,9 @@ P_INTERNAL_LINK_DEF = re.compile(r'^\[([^\]]+)\]:\s*(.+)')
 # Pattern to match {% include ... %} statements
 P_INTERNAL_INCLUDE_LINK = re.compile(r'^{% include ([^ ]*) %}$')
 
+# Pattern to match image-only and link-only lines
+P_LINK_IMAGE_LINE = re.compile("^[> ]*(!?)\[([^]]+)\][([]([^)]+)[])][ ]*$")
+
 # What kinds of blockquotes are allowed?
 KNOWN_BLOCKQUOTES = {
     'callout',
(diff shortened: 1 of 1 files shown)


fmichonneau (approved):

fmichonneau:
Thanks Maxim!

maxim-belkin:
Thanks, François!

@jhlegarreta, could you please confirm that after this change `make lesson-check-all` doesn't complain about the long lines that contain images or links only?

jhlegarreta:
The regex does not seem straightforward ! Thanks both :+1:.

> @jhlegarreta, could you please confirm that after this change make lesson-check-all doesn't complain about the long lines that contain images or links only?

Will give it a try when I find the time (will try during the weekend) and will report back.

jhlegarreta:
Unfortunately, it does not look like it solved the issue. In fact, the number of warnings has increased, e.g.:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2531577900?check_suite_focus=true#step:16:20
with respect to an immediately previous version, e.g.
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2528767166?check_suite_focus=true#step:16:20

In the previous version, link-only lines (e.g. line 190 at the pointed file - `__episodes/constrained_spherical_deconvolution.md`) were not triggering warnings, and now they are. Meanwhile, the lines that made me open (e.g. line 458) the issue are still raising the warnings.

maxim-belkin:
Oh, I see what the problem is. The pattern I added doesn't allow `{:class ...}`at the end. I'll fix that.

maxim-belkin:
So, I found a few things that you use in your lesson that I didn't take into account when I was working on this pattern:

1. You use links in headings
2. You use `{:class=...}` customizations in link/image-only lines
3. You use `\` at the end of the image/link-only lines

So, #597 should address all of these.

PR #596, opened 2021-05-07
bin/util.py: Factor out reporter class. Define __all__

maxim-belkin:
The PR factors out the Reporter class into its own file. 
This should make `util.py` a bit easier to parse.
\---
The second change (defining `__all__` in `util.py`) allows to use `from util import *` to import the elements defined in `__all__`.

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -8,8 +8,8 @@ import glob
 import re
 from argparse import ArgumentParser
 
-from util import (Reporter, read_markdown, load_yaml, check_unwanted_files,
-                  require)
+from util import *
+from reporter import Reporter
 
 __version__ = '0.3'
 
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -8,6 +8,8 @@ import glob
 import re
 from argparse import ArgumentParser
 
+# This uses the `__all__` list in `util.py` to determine what objects to import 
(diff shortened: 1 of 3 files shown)


maxim-belkin:
@zkamvar @fmichonneau, OK for me to merge this one? (are there any questions or concerns?) I'd like to restructure our Python code a little bit and maybe convert it into a Python module (which we can `pip install carpentries-<something>`) down the road.

zkamvar:
I am all for splitting apart large scripts in to smaller modules! TIL about the `__all__` list in python: https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

bin/lesson_check.py
```diff
@@ -8,8 +8,8 @@
 import re
 from argparse import ArgumentParser
 
-from util import (Reporter, read_markdown, load_yaml, check_unwanted_files,
-                  require)
+from util import *
```
zkamvar, line 13: For future-me, who is not terribly well-versed in Python, this will be helpful to remind me of what `import *` is doing here.

```suggestion
# This uses the `__all__` list in `util.py` to determine what objects to import 
# see https://docs.python.org/3/tutorial/modules.html#importing-from-a-package
from util import *
```

zkamvar:
Thank you, Maxim!

maxim-belkin:
Thank you, Zhian!

PR #597, opened 2021-05-09
lesson_check.py: harden single-line image/link pattern

### LINKED ISSUE #591: Allow skipping line length checks for link-only lines

maxim-belkin:
This change hardens the pattern that matches single-line
image or link:

1. It extends the pattern to be matched in a heading
2. It allows the line to contain {: ...} customizations
3. It allows the line to end with \

Fixes carpentries/styles#591

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -57,7 +57,16 @@ P_INTERNAL_LINK_DEF = re.compile(r'^\[([^\]]+)\]:\s*(.+)')
 P_INTERNAL_INCLUDE_LINK = re.compile(r'^{% include ([^ ]*) %}$')
 
 # Pattern to match image-only and link-only lines
-P_LINK_IMAGE_LINE = re.compile("^[> ]*(!?)\[([^]]+)\][([]([^)]+)[])][ ]*$")
+P_LINK_IMAGE_LINE = re.compile(r'''
+    [> #]*        # any number of '>', '#', and spaces
+    !?            # ! or nothing
+    \[[^]]+\]     # [any text]
+    [([]          # ( or [
+    [^])]+        # 1+ characters that are neither ] nor )
+    [])]          # ] or )
+    (?:{:[^}]+})? # {:any text} or nothing
+    [ ]*          # any number of spaces
+    \\?$          # \ or nothing + end of line''', re.VERBOSE)
 
 # What kinds of blockquotes are allowed?
 KNOWN_BLOCKQUOTES = {


maxim-belkin:
@jhlegarreta, here is the list of "long lines" that I get for your lesson with this change:

```
./CONTRIBUTING.md: Line(s) too long: 3, 114, 126
./README.md: Line(s) too long: 2, 3, 4, 5, 6, 7, 10, 14, 17, 18, 23, 25, 59, 60, 61, 64, 65
./_episodes/constrained_spherical_deconvolution.md: Line(s) too long: 69, 77
./_episodes/deterministic_tractography.md: Line(s) too long: 44, 61, 267, 272, 274, 283
./_episodes/diffusion_tensor_imaging.md: Line(s) too long: 32, 48, 52, 53, 56, 121, 166, 237, 320
./_episodes/preprocessing.md: Line(s) too long: 137, 175, 193, 243, 245, 270, 271, 273, 347, 368, 392
./_episodes/probabilistic_tractography.md: Line(s) too long: 73, 92
./_extras/discuss.md: Line(s) too long: 76
./_extras/guide.md: Line(s) too long: 30
./index.md: Line(s) too long: 30
```

The lines you were concerned about seems to be no longer reported as "too long".

Line 190 in `_episodes/constrained_spherical_deconvolution.md` was triggering the "Line too long" warning with the previous pattern because it wasn't matching against it due to `{:class=...}` and `\` at the end of the line. Before that, this line was skipped by this check because it begins with an `!`.

Line 458 in the same file was triggering the "Line too long" warning again due to `{:class=...}` and `\` at the end of the line not letting the previous pattern to match it.

maxim-belkin:
I'm going to merge this one as all the checks have passed and I'm pretty confident it addresses the issues reported. 

@jhlegarreta, let me know if there is some other extra syntax that you use (or would like to use) for links/images that I haven't captured in this regex.

jhlegarreta:
@maxim-belkin thanks for the explanation in the commit message, and your time and effort !

The new changeset does fix the lines I pointed to :+1:. The relevant part of my `make lesson-chell-all` output is exactly the same as the one you posted.

There are still three instances that are not handled correctly:
`./_episodes/diffusion_tensor_imaging.md: 48, 56, 166`

I am not sure in which ways those lines are different from other instances. I have noticed that they have a trailing whitespace, and they only contain a link, but apart from that I see no other noticeable different to the many similar instances in that file.

Some other lines, e.g. `37`, do have trailing whitespaces, are link-only lines within a paragraph, exceed the 100 character limit, and are not triggering a warning.

I grant that trailing whitespaces should always be trimmed, and that we should configure our tools to do that. If you conclude that the reason behind the above lines not being handled correctly are the trailing whitespaces, I'll remove them all from the file. However, note that I went ahead and removed them locally (only for the pointed lines), then passed the check again and the warnings are still there.

So there is still something that is elusive to us.

maxim-belkin:
> There are still three instances that are not handled correctly:
> `./_episodes/diffusion_tensor_imaging.md: 48, 56, 166`

All of these lines are images and they have the following parts that don't match the pattern:
 - Line 48 has a dot at the end.
 - Line 56 has a dot at the end.
 - Line 166 has a comma at the end.

What's interesting (at least, I've never seen this done before) is that you treat images' alt text as part of the corresponding paragraphs. I think that due to accessibility considerations paragraphs should refer to images rather than use them this way.

> Some other lines, e.g. `37`, do have trailing whitespaces, are link-only lines within a paragraph, exceed the 100 character limit, and are not triggering a warning.

Line 37 is an image-only line, doesn't have neither a period nor comma at the end, so it matches the pattern and is, therefore, skipped by this check.

> I grant that trailing whitespaces should always be trimmed

Two trailing whitespaces force line breaks in Markdown - see 740773a (PR https://github.com/carpentries/styles/pull/522), so you should keep them, if that's what you use them for.

jhlegarreta:
Again, thanks for having had a look at this @maxim-belkin.

> All of these lines are images and they have the following parts that don't match the pattern:
    Line 48 has a dot at the end.
    Line 56 has a dot at the end.
    Line 166 has a comma at the end.

Thanks for identifying the reason. As I see it, having a punctuation mark at the end should not be a reason to miss these cases. But let me know if I'm wrong.

> What's interesting (at least, I've never seen this done before) is that you treat images' alt text as part of the corresponding paragraphs. I think that due to accessibility considerations paragraphs should refer to images rather than use them this way.

Not sure if "treating images' alt text as part of the corresponding paragraphs" was intended. I do not see its effects or how this can be done otherwise. Alt texts were introduced so that all images, even the inline ones, have a caption when being listed in
https://carpentries-incubator.github.io/SDC-BIDS-dMRI/figures/index.html

So not sure how this comment and the above are intertwined so that we can fix the cases.

> Two trailing whitespaces force line breaks in Markdown - see 740773a (PR #522), so you should keep them, if that's what you use them for.

The intention was not to force line breaks; maybe it was just the way the contributor added the text. But I did not know about the two trailing whitespaces. Interesting. Thanks.

maxim-belkin:
> As I see it, having a punctuation mark at the end should not be a reason to miss these cases.

Currently, images in the lessons are displayed as block elements (line 256 below), meaning that they start on new lines and take up the whole width of the page:

https://github.com/carpentries/styles/blob/af7efa5763f50f1b380ce0348349e68a4cdab56e/assets/css/lesson.scss#L255-L259

Because of that, text that follows immediately after an image in Markdown ends up on a new line in HTML, see the screenshot below:
<details style="border: solid 1px black;"><summary>screenshot</summary>

rendered version of the "Diffusion Tensor Imaging" episode of the "Introduction to dMRI" lesson
</details>

So, all of that was to say that our lessons don't currently support inline images. But they could and, probably, should. I'll submit a PR proposing this and another PR to allow punctuation marks after links and images because it seems logical to be able to use punctuation marks after the inline images.

jhlegarreta:
Thanks for having had a close look at this :100:.

> Currently, images in the lessons are displayed as block elements (line 256 below), meaning that they start on new lines and take up the whole width of the page:
> Because of that, text that follows immediately after an image in Markdown ends up on a new line in HTML, see the screenshot below:

I see. Yes, I had noticed that what we pretended to be inline images were in fact not so. I did not investigate further, though. Now I know the reason :+1:.

> So, all of that was to say that our lessons don't currently support inline images. But they could and, probably, should. I'll submit a PR proposing this and another PR to allow punctuation marks after links and images because it seems logical to be able to use punctuation marks after the inline images.

Excellent ! Both endeavors make sense to me as separate PRs. They are by no means urgent. At least we now know better the reasons behind.

maxim-belkin:
@jhlegarreta, I submitted PR #599 that allows up to 3 non-word characters to appear on the same line before and after a link or an image. I chose 3 characters as a middle ground between "too few" and "too many". I think it's OK to allow non-word characters in the beginning of the image because I noticed that in the carpentries-incubator/SDC-BIDS-dMRI lesson there is at least one case where an image is surrounded by parentheses.

jhlegarreta:
Thanks @maxim-belkin !

Looks like our `diffusion_tensor_imaging.md` episode gets a nice benefit from the PR:
```
./_episodes/diffusion_tensor_imaging.md: Line(s) too long: 32, 53, 121, 320
```
vs. without the changes in the PR:
```
./_episodes/diffusion_tensor_imaging.md: Line(s) too long: 32, 48, 52, 53, 56, 121, 166, 237, 320
```

PR #599, opened 2021-05-11
lesson_check.py: relax P_LINK_IMAGE_LINE pattern

maxim-belkin:
This PR allows up to 3 non-word (`\W` in Python's `re`-speak) characters
in the beginning and end of the pattern that matches links and images.
This is to allow lesson developers place punctuation marks, parentheses,
or other symbols before and after the link or image on the same line in
Markdown.


maxim-belkin:
Merging, as it seems to have the desired effect (based on @jhlegarreta's [comment in #597][1]).

[1]: https://github.com/carpentries/styles/pull/597#issuecomment-839379061

PR #601, opened 2021-05-14
Don't force hostname into relative_root_path

unode:
When using platforms such as https://gitpod.io the content generated via `make serve` has `site.url` set to `http://localhost:4000` but is served via a web proxy with an arbitrary hostname (e.g. `https://4000-ivory-camel-r25nr08e.ws-eu01.gitpod.io`).

The existing code hardcodes the hostname breaking URLs served through said proxy.

I can't think of any use-case that would break if the hostname is absent, so this PR removes `site.url` and uses only `site.baseurl`.
I tested the use-cases described in #569 and #578 and all seems to work as expected.

### DIFF
--- a/_includes/base_path.html
+++ b/_includes/base_path.html
@@ -7,7 +7,7 @@ When we start a local server using `jekyll serve`,
 'site.url' is set to 'http://localhost:4000' and
 'site.baseurl' is empty.
 
-In both of the above cases we set 'relative_root_path' to 'site.url + site.baseurl'.
+In both of the above cases we set 'relative_root_path' to 'site.baseurl'.
 
 When we build a website locally with `jekyll build`,
 both 'site.url' and 'site.baseurl' are empty.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thank you, @unode! This is really awesome! I don't use gitpod.io so I didn't realize there is such a use case. In my tests, your solution didn't break `bundle exec jekyll serve` nor GitHub Pages, so it looks good to me. Could you please update the comment on line 10 that starts with `In both of the above cases...` to reflect the new state of things?

unode:
Comment is now consistent with the code.

maxim-belkin (approved):

maxim-belkin:
Thank you, Renato! Great work! 🎉  I'm excited that we've managed to make `relative_root_path` work correctly in all of these cases!

unode:
Cheers Maxim! It's looking great indeed :+1:

PR #606, opened 2021-06-07
Automatically add deep anchor links using AnchorJS

maxim-belkin:
I propose to use [AnchorJS](https://www.bryanbraun.com/anchorjs) to automatically add (deep) anchor links. Here is a quick demo:

https://user-images.githubusercontent.com/13123663/120898940-d1599a80-c5f2-11eb-9f8b-225a5e9af616.mov


zkamvar:
I think adding anchors is a great idea. Is there a reason to include the script as opposed to loading it from the CDN?

maxim-belkin:
> Is there a reason to include the script as opposed to loading it from the CDN?

Only to make offline lesson development possible.

fmichonneau (approved):

maxim-belkin:
Thanks for the approval, @fmichonneau.
I tested this change locally, and it seems to work as intended (it doesn't mess up the links that kramdown creates).

PR #608, opened 2021-06-24
Makefile: docker-serve target: ensure Docker is installed

maxim-belkin:
Ensure that Docker is installed prior to executing the `docker-serve` target. Emit a clear message in case Docker is not available.

**Before**

```
$ make docker-serve
make: docker: No such file or directory
make: *** [docker-serve] Error 1
```

**After**

```
$ make docker-serve
Makefile:57: *** Your system does not appear to have Docker installed.  Stop.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ JEKYLL=bundle config set --local path .vendor/bundle && bundle install && bundle
 PARSER=bin/markdown_ast.rb
 DST=_site
 
+# Find Docker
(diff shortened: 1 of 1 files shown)


maxim-belkin:
I updated implementation from 507cdad to f7e6c1a because, unlike Python which is required by several targets, Docker is required by `docker-serve` only, so we can check that Docker is installed inside the `docker-serve` target only.

fmichonneau (approved):

PR #616, opened 2021-07-12
lesson-check.py: read `config.yml` only once

maxim-belkin:
1. Read `config.yml` file only once and store the contents in a global variable `CONFIG`. We're currently reading this file twice.
2. Detect lesson life cycle in the `main()` function instead of making `check_config()` function return it.
3. Replace `using_remote_theme` function with a test that checks that `remote_theme` keyword is present in `_config.yml`
4. CheckEpisode class: `check` method: move "remote theme test" inside the `check_reference_inclusion`.
5. main: move "remote theme test" inside the `read_references`

### DIFF
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -9,7 +9,7 @@ import re
 import sys
 from argparse import ArgumentParser
 
-# This uses the `__all__` list in `util.py` to determine what objects to import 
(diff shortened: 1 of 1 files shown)


maxim-belkin:
All checks have passed, so merging. There should be no effect other than reading the `_config.yml` file only once.

PR #635, opened 2022-01-21
update e-mail address

brownsarahm:
this is to match updates made to the checkout instructions on the instructor training curriculum: https://github.com/carpentries/instructor-training/pull/1300


zkamvar:
Thank you!

PR #636, opened 2022-02-21
Use a pixelated img rendering instead of blurred default

unode:
Workaround to a CSS resize blur rendering problem affecting Google Chrome/Chromium
when the resolution of the image is larger than the image view

See https://stackoverflow.com/a/69156216 for discussion and
https://github.com/carpentries-incubator/jekyll-pages-novice/issues/274 where the issue was initially reported.

### DIFF
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -321,6 +321,9 @@ span.fold-unfold {
   opacity: 0.5;
 }
 
+img {
+  image-rendering: pixelated;
+}
 
 p.image-with-shadow img,
 img.image-with-shadow {
--- a/assets/css/lesson.scss
+++ b/assets/css/lesson.scss
@@ -321,6 +321,12 @@ span.fold-unfold {
   opacity: 0.5;
 }
 
+/*
+  Force a pixelated rendering of images when original resolution is bigger than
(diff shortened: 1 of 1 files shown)


maxim-belkin:
The change LGTM, but I think it is worth adding a comment explaining what this change is trying to achieve and, perhaps, linking to relevant issues and StackOverflow articles.

assets/css/lesson.scss
```diff
@@ -321,6 +321,9 @@ span.fold-unfold {
   opacity: 0.5;
 }
 
+img {
```
unode, line 330:
```suggestion
/*
  Force a pixelated rendering of images when original resolution is bigger than
  the <img> viewport to avoid a blurring effect that makes it difficult to read the image.
  This issue affects primarily Chrome based browsers (as of 2021).
  See https://github.com/carpentries/styles/pull/636 for the original report and related links 
*/
img {
```

unode: @maxim-belkin does this look reasonable?

maxim-belkin: > @maxim-belkin does this look reasonable?

Yes, thank you!

fmichonneau (approved):

fmichonneau:
Thank you both!

PR #639, opened 2022-10-05
modify transition workflow/variables

zkamvar:
1. The transtion headers will now have a button that can close them
   (though its not permenent)
2. The third step of the transition will no longer be coded into the
   yaml here because the third step of the transition will use the
   workbench version as the default version. This means that no redirect
   will be needed.
3. All transition dates will be specified
4. Language has been updated to not refer to "Jekyll" as it may not be
   immediately obvious that Jekyll was used to build the site.

Screenshots:

With `transition-step-1`:

screenshot of warning banner that says there is a snapshot of this lesson being tested on The Carpentries Workbench and that it will transition over in 2023

With `transition-step-2`:

screenshot of danger banner that says this lesson is a snapshot from November and a newer version is being tested on The Carpentries Workbench and that it will transition over in 2023

### DIFF
--- a/_includes/check_transition_variables.html
+++ b/_includes/check_transition_variables.html
@@ -1,12 +1,5 @@
 {% assign date = include.need_transition_date %}
 
-{% unless site.transition_url %}
-<div class="alert alert-danger">
-  you need to specify the variable <code>transition_url</code>
-  in <code>_config.yml</code>.
-</div>
-{% endunless %}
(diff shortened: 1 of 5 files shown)


zkamvar:
Note: windows is failing because of some esoteric bundler error. Given that we are on track to replace this infrastructure, these failures are non-critical.

PR #643, opened 2022-10-20
Fix workflows 641

### LINKED ISSUE #641: [workflows] Unable to resolve action r-lib/actions@master
## Solution

NOTE: this issue has been fixed for repositories within The Carpentries organisations. The instructions below are for anyone who has a repository outside The Carpentries.

### Option 1: Replace Files

**If you have a lesson or workshop repository outside of The Carpentries and are experiencing this issue** check to make sure you have the following:

 - [ ] Is your repository outside of The Carpentries?
 - [ ] Does your repository have the folder `.github/workflows`?
 - [ ] Do you have the either or both "website.yml" and/or "template.yml" files inside `.github/workflows`?

If you have checked all three of the boxes above, replace these files in your repository with [the workflow files from the styles repository](https://github.com/carpentries/styles/tree/gh-pages/.github/workflows). 

### Option 2: Modify Files

Alternatively, you can replace the individual steps in your YAML files.

Both of these actions calls need to be updated to use

```yaml
        uses: r-lib/actions/setup-r@v2
        with:
          use-public-rspm: true
          install-r: false
```

There is also a [recently deprecated workflow function](https://github.com/carpentries/styles/issues/642) that needs one line to be changed in the workflows: 

```diff
-          echo "::set-output name=count::$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})"
+          echo "count=$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})" >> $GITHUB_OUTPUT
```

## Details of problem

Originally reported in https://github.com/datacarpentry/spreadsheet-ecology-lesson/pull/317#issuecomment-1285542583,

https://github.com/carpentries/styles/blob/6b8bbad1187c34b7ed321ed5d4fea0866cddadfc/.github/workflows/website.yml#L51-L53 and 
https://github.com/carpentries/styles/blob/6b8bbad1187c34b7ed321ed5d4fea0866cddadfc/.github/workflows/template.yml#L134-L136

are failing because `r-lib/actions/setup-r@master` no longer exists.

## Impact

This impacts all repositories that were built with the styles template or the remote theme. 

 - In all repositories, all pull requests will fail checks.
 - In repositories that use R Markdown, the lessons will fail to build.
 - Good news: In all other repositories, the **lessons will continue to build**. 

Because this is an urgent update that involves six lines of workflow code that is not _normally_ touched by the maintainers, the deployment of this fix is better handled through mass pull requests that target these specific items in the workflow YAML. The reasoning for this is that we have estimated previously that manually making the pull requests to merge styles into these repositories  would take roughly a day of full time work, assuming nothing goes wrong. Figuring out how to write GitHub API calls to fix this solution quickly is a much better use of time even if it creates merge conflicts with styles down the line.

### LINKED ISSUE #642: [workflows] set-output deprecated
The set-output command is deprecated in favour of the `$GITHUB_OUTPUT` variable: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

zkamvar:
This updates workflow steps to ensure that we are using the correct setup-r action and that we are using updated output setting commands. 

This will address #641 and fix #642

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -127,7 +127,7 @@ jobs:
         id: check-rmd
         working-directory: lesson
         run: |
-          echo "::set-output name=count::$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})"
+          echo "count=$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})" >> $GITHUB_OUTPUT
 
       - name: Set up R
         if: steps.check-rmd.outputs.count != 0
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -131,9 +131,10 @@ jobs:
 
       - name: Set up R
         if: steps.check-rmd.outputs.count != 0
-        uses: r-lib/actions/setup-r@master
+        uses: r-lib/actions/setup-r@v2
         with:
-          r-version: 'release'
+          use-public-rspm: true
+          install-r: false
 
       - name: Install needed packages
         if: steps.check-rmd.outputs.count != 0
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -48,9 +48,10 @@ jobs:
 
       - name: Set up R
         if: steps.check-rmd.outputs.count != 0
-        uses: r-lib/actions/setup-r@master
+        uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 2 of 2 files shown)


zkamvar:
The Geospatial lesson is currently failing only on MacOS, but that is due to a weirdness with "stringr" that I do not want to address right now.

